PR #20, opened 2015-05-19
Use correct path for gulp

cphyc:


package.json
```diff
@@ -4,8 +4,8 @@
   "description": "A Framework that enables Real-Time Collaboration on arbitrary data structures.",
   "main": "./build/node/y.js",
   "scripts": {
-    "prepublish": "./node_module/.bin/gulp build_node",
```
DadaMonad: it's actually the same thing, but ok

PR #24, opened 2015-06-05
Add example of integration using easyrtc and angular

cphyc:


DadaMonad:
Hey, this is awesome!!

Angular seems to be a really good fit for Yjs. But could you put this in a Gist instead? I will link to it from the wiki. I want to provide running examples in this repository only.

Thanks for the contrib :-)

cphyc:
Sure ! See here: https://gist.github.com/03f153ebf4ccf265805e.git

DadaMonad:
Thx :shipit:

I put it in the wiki (Connector-howto Conclusion).

PR #27, opened 2015-07-06
Add getNext and operations on references

cphyc:


lib/Operations/Structured.coffee
```diff
@@ -155,6 +155,19 @@ module.exports = ()->
     getFirstOperation: ()->
       @beginning.next_cl
 
+    # get the next non-deleted operation
+    getNext: (start)->
```
DadaMonad: `start.getNext()` yields the same result. You can also call `start.getNext(i)` to get the i-th non-deleted operation

lib/Operations/Structured.coffee
```diff
@@ -232,6 +245,14 @@ module.exports = ()->
     push: (content)->
       @insertAfter @end.prev_cl, [content]
 
+    insertAfterHelper: (root, content)->
```
DadaMonad: What does this method do? What is `content.bt`, or `root.bt`?

cphyc: Oops, shouldn't be here. I'm updating the PR, until it's mature, I'm cancelling it.

PR #29, opened 2015-07-28
Improve performance

cphyc:


DadaMonad:
This looks actually really nice!

I can't compile the source code. It says: RBTreeByIndex is not defined. I assume it is because of `RBTReeByIndex = require 'bintrees/lib/rbtree_by_index' is not defined.

Can you also make sure that the tests of Y.Text and Y.Array succeed? (It is the same build system we use in y-richtext: `gulp mocha` should succeed.

It would be nice if you could put your rbtree implementation on npm.org. Then ther would be some kind of version control, and the build also succeeds when you change something in your repository.

For maximum performance improvement, you should also change `Insert.prototype.get(Next+Prev)` which is something like your `ListManager.prototype.get(Next+Prev)NonDeleted`. Better: delete one of them. Furthermore, I don't understand why there is a tree for the complete list too. Don't you think that this is a bit overkill? - it is already saved by the operation (`*.next_cl`)

`

cphyc:
Hi,
I'm away until the 16th so I won't fix it until then, at least!
Cheers (from Georgia)

On August 6, 2015 4:17:01 PM GMT+04:00, Kevin Jahns <notifications@github.com> wrote:
>This looks actually really nice!
>
>I can't compile the source code. It says: RBTreeByIndex is not defined.
>I assume it is because of `RBTReeByIndex = require
>'bintrees/lib/rbtree_by_index' is not defined.
>
>Can you also make sure that the tests of Y.Text and Y.Array succeed?
>(It is the same build system we use in y-richtext: `gulp mocha` should
>succeed.
>
>It would be nice if you could put your rbtree implementation on
>npm.org. Then ther would be some kind of version control, and the build
>also succeeds when you change something in your repository.
>
>For maximum performance improvement, you should also change
>`Insert.prototype.get(Next+Prev)` which is something like your
>`ListManager.prototype.get(Next+Prev)NonDeleted`. Better: delete one of
>them. Furthermore, I don't understand why there is a tree for the
>complete list too. Don't you think that this is a bit overkill? - it is
>already saved by the operation (`*.next_cl`)
>
>`
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/y-js/yjs/pull/29#issuecomment-128345215

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

DadaMonad:
Don't worry. Have fun in Georgia!

PR #110, opened 2018-04-12
Add local rollup dependency to build

AquiGorka:
You probably have it installed globally.


dmonad:
Thanks for the PR! I just fixed it.

PR #138, opened 2019-01-16
updating YArray's iterator to iterate Types correctly 

calibr:
Looks like it was missing a switching to the next item after processing a Type item, so once iterator met a Type it stuck on it. If more tests or something else is required I will be happy to do that.


dmonad:
This is great. Thanks for including a test too!

PR #172, opened 2019-10-05
Fixing Y.Map's documentation of forEach

### LINKED ISSUE #171: Inconsistent order of arguments in Y.Map vs. documentation
The README documentation says the Y.Map type exposes a `f(key, value)` iterator, however, the source code in https://github.com/yjs/yjs/blob/ece4841b5c503ae4bac89c3d451d182c9ba3d958/src/types/YMap.js#L150
apparently does `f(value, key)`. Am I missing something!? The documentation for Y.Array seems correct. If you confirm, I am more than happy to offer a PR. :grinning:

istvank:
fixes #171 As always, it's an honor to submit a PR! 🐒 There was also a missing dot in the Y.XmlFragment title.

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,7 +235,8 @@ or any of its children.
 Copies the <code>[key,value]</code> pairs of this YMap to a new Object.It
 transforms all child types to JSON using their <code>toJSON</code> method.
     </dd>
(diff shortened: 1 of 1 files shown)


dmonad:
Thanks for the help @istvank! Always happy to receive a PR from you :)

PR #174, opened 2019-10-15
Fix attrs loop in yXmlText

boschDev:
The element attribute loop use the index of the nodes array. Updated to use the index of the attributes array. The problem is that by the first node the code will only apply the first attribute and for the second node only the second attribute. In the other yXml types are fine.


dmonad:
Of course, thank you for noticing!

PR #176, opened 2019-12-10
fix setting Doc.gc value in Doc's constructor

calibr:
now it is impossible to disable gc in a document, because Doc's constructor always sets gc to true regardless of config


dmonad:
Right, I don't know what I was thinking there. Thanks for bringing that up. I fixed the issue in another PR to also create proper typings.

PR #179, opened 2020-01-22
basic Y.Text toJSON returns {unformatted:string}

garth:
This avoids text nodes in nested structures returning undefined when toJSON is called by a parent.

### DIFF
--- a/src/types/YText.js
+++ b/src/types/YText.js
@@ -669,6 +669,17 @@ export class YText extends AbstractType {
   }
 
   /**
+   * Returns the unformatted string representation of this YText type.
+   *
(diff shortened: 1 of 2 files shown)


dmonad:
Thanks @garth ! You are right, toJSON shouldn't return undefined. But I think it should return the plain string instead (`this.toString()`). If you make that change I'd be happy to merge your PR.

garth:
@dmonad, I wasn't sure about if if should be a simple string or json. Updated as you suggested.

dmonad:
Thanks for the PR! Will publish a new version in a bit..

PR #194, opened 2020-04-27
Add example of combining providers to README.md

mohe2015:
Open questions:
- is the place fine? I don't really like it but couldn't find a better one.
- is the example itself fine?


mohe2015:
What do you think about this? Is it fine like this, should I only leave [line 140-160](https://github.com/yjs/yjs/pull/194/files#diff-04c6e90faac2675aa89e2176d2eec7d8R140-R160) in? Anything else?

dmonad:
Hey @mohe2015 , thanks for the contribution! I updated the demo a bit and published it.

PR #199, opened 2020-05-12
Enhance Y.Array insertions to behave like normal arrays

Mansehej:
Good day @dmonad.

I have tried to make Y.Array functions behave a bit more like normal arrays do in JavaScript.
Here are the changes I have made:
- Add unshift(Array<T>) operation which preppends the content to the Y.Array.
- insert, push, and unshift now return the length of the updated content.

I hope these changes would be helpful, as people can expect Y.Array behave a bit more like normal arrays.

I also have a question:
I was thinking of implementing pop() and shift(). However, considering that push and unshift take an array as an argument, will it not create inconsistency if pop() and shift() only delete and return the last element? I was thinking of taking an argument of length (ex. pop(3) to delete and return the 3 topmost elements of the Y.Array). What do you think?

Regards,
Mansehej.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,8 @@ position 0.
     </dd>
     <b><code>push(Array&lt;Object|boolean|Array|string|number|Uint8Array|Y.Type&gt;)</code></b>
     <dd></dd>
+    <b><code>unshift(Array&lt;Object|boolean|Array|string|number|Uint8Array|Y.Type&gt;)</code></b>
+    <dd></dd>
     <b><code>delete(index:number, length:number)</code></b>
     <dd></dd>
     <b><code>get(index:number)</code></b>
--- a/src/types/YArray.js
+++ b/src/types/YArray.js
@@ -101,6 +101,7 @@ export class YArray extends AbstractType {
    *
    * @param {number} index The index to insert content at.
    * @param {Array<T>} content The array of content
+   * @return {number} The upated YArray length
    */
   insert (index, content) {
     if (this.doc !== null) {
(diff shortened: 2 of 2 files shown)


Mansehej:
Hey, could you please review this once, @dmonad?

dmonad:
Hey @Mansehej,

One of the great opportunities of this project is that it does Arrays don't have to behave like JavaScript Arrays.

I was actually always thinking about making all the methods in Yjs chainable. I.e. Every method returns its own instance. This would allow for some cool tricks:

```js
array.insert(5, [1, 2, 3]).delete(0, 2).toArray()
```

But even more useful on Y.Maps:

```js
map.set('key1', 1)
       .set('key2', 2)
       .delete('key3')
```

I really appreciate your contribution, but the return values are reserved for chainability. I would gladly accept the unshift contribution though.

Mansehej:
Oh that's a great idea! I'll make another PR for the unshift commit.

Also, mind if I try my hand at chaining the functions?

dmonad:
Sure, I'd be happy to receive a PR for that :)

Please make two separate PRs. One for the unshift change, and another one for the chaining functions (this one might receive more feedback).

All "setter" methods of the Yjs types should return the type itself. Please also add the jsdoc type information. In the readme it should return the type itself:

```js
<b><code>unshift(Array&lt;Object|boolean|Array|string|number|Uint8Array|Y.Type&gt;):Y.Array</code></b>
```

Looking forward to your PR. I'll close this one.

PR #212, opened 2020-06-12
Y.Map support for forEach, get, has on prelimContent

### LINKED ISSUE #207: Y.Map get does not return prelim content
When a Y.Map has a key/value set, I expect it to return the value I just set:

```
const ymap = new Y.Map()
ymap.set('value', 0)
ymap.get('value') // -> undefined
```

I believe this is caused by the above Y.Map not being backed by a Y.Doc yet, so its 'value' is set as prelim content. When I initialize a Doc it behaves as I expect:

```
const ydoc = new Y.Doc()
const ymap = ydoc.getMap('mymap')
ymap.set('value', 0)
ymap.get('value') // -> 0
```

Should both of these cases return 0?

**Environment Information**
 - Firefox 77.0.1
 - Yjs 13.0.4

canadaduane:
Fixes #207 and adds `has` and `forEach` as well.

### DIFF
--- a/src/types/YMap.js
+++ b/src/types/YMap.js
@@ -155,18 +155,21 @@ export class YMap extends AbstractType {
    * Executes a provided function on once on every key-value pair.
    *
    * @param {function(T,string,YMap<T>):void} f A function to execute on every element of this YArray.
+   * @returns {Object<string,T>}
    */
   forEach (f) {
-    /**
-     * @type {Object<string,T>}
-     */
-    const map = {}
-    for (const [key, item] of this._map) {
-      if (!item.deleted) {
-        f(item.content.getContent()[item.length - 1], key, this)
+    if (this.doc !== null) {
+      for (const [key, item] of this._map) {
+        if (!item.deleted) {
+          f(item.content.getContent()[item.length - 1], key, this)
+        }
+      }
+    } else {
+      for (const [key, item] of /** @type {Map<string, any>} */ (this._prelimContent)) {
+        f(item, key, this)
       }
     }
-    return map
+    return {}
   }
 
   /**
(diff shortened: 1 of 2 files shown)


dmonad:
Thanks for the PR!, but I don't think it is a good idea to merge it into a stable release. As I explained, this is the intended behavior and fixing it partially will almost definitely bring up some issues in other projects (especially in editor bindings).

I agree that this could be easily integrated for the Y.Map type. But it won't be that easy for the Y.Text type. I dislike implementing too many inconsistencies in the types. It is understandable to say "Types don't have any content and don't fire event until they are initialized.". But if some of them have content, they might expect that there is a bug in Yjs.

The other problem are the observers. The whole observer pattern doesn't make sense if we start supporting getters before initialization:

```js
const ytext = new Y.Text()
ytext.observe(event => console.log('event fired'))
ytext.insert(0, ''some content") // no event occurs, because it is not hooked to a Yjs instance
ytext.toString() // => 'some content'  - but content already exists? - before observer is fired?
map.set('somekey', ytext) // => 'event fired'
```

This might also break some of the existing data bindings. A data binding usually retrieves the current content from the type and then listens to changes. So using the above example, you'd end up with double the content, because the initial content is added again after the content is added to the Yjs document:

```js
// how a data binding works:
codemirror.setContent(ytext.toString())
ytext.observe(event => { // this would fire again after the content is added to ytext
  codemirror.applyDelta(event.changes.delta)
})
```

In the next stable release, I'd like to have types to be independent from the Y.Doc.  For now I suggest that you use the Yjs types only in combination with a Y.Doc and only initialize content before you bind it to a Y.Doc.

PR #224, opened 2020-07-23
Document the  doc.toJSON method

### LINKED ISSUE #222: Document toJSON API
Hi!  I'm excited to dip my feet back into the land of CRDTs after a few years of following from a distance.

This issue is just a quick question.  I noticed that the `.toJSON()` API was added a few days ago, but hasn't been documented yet.  Is that a feature you would like to be documented?

I wouldn't mind proposing a diff, if you think it could be helpful.  :)

ajhyndman:
Here's a first draft!

I reasoned that it's most valuable to just know that this method exists, but I'm not sure if there's anything extra you wanted to communicate.

Fixes #222


dmonad:
This is great, thanks :-)

Issue #159, opened 2019-06-27
restoring document to a specific state


calibr:
I see that new yjs is coming with Snapshots, will they be able to restore the whole document to a specific state based on StateSet and DeleteSet ?

What I did in the earlier version was essentialy packing the document into a syncStep2 message to specific state and then feeding this message to a new document.

```
function backDocToState(doc, state) {
  const { ss, ds } = state
  const needState = new Map<number, number>(ss)
  const encoder = encoding.createEncoder()
  encoding.writeVarUint(encoder, syncProtocol.messageYjsSyncStep2)
  const lenPos = encoding.length(encoder)
  encoding.writeUint32(encoder, 0)
  let len = 0
  const tempStructs = []
  for (let user of needState.keys()) {
    let clock = needState.get(user) || 0
    if (user !== ID.RootFakeUserID) {
      const minBound = ID.createID(user, 0)
      doc.os.iterate(minBound, ID.createID(user, clock - 1), struct => {
        tempStructs.push(struct._id, struct._length, struct._content)
        if((struct._id.clock + struct._length - 1) > clock) {
          const lenDif = struct._id.clock + struct._length - clock - 1
          struct._splitAt(doc, lenDif)
        }
        struct._toBinary(encoder)
        len++
      })
    }
  }
  encoding.setUint32(encoder, lenPos, len)

  const tmpDeleteStore = new Y.DeleteStore()
  for(let [userId, clock, length] of ds) {
    tmpDeleteStore.mark(ID.createID(userId, clock), length)
  }

  Y.writeDeleteStore(encoder, tmpDeleteStore)

  const buf = encoding.toBuffer(encoder)
  const decoder = decoding.createDecoder(buf)

  const newDoc = new Y.Y()
  newDoc.transact(() => {
    syncProtocol.readSyncMessage(decoder, encoding.createEncoder(), newDoc)
  })
}
```

Probably I can do the same in the current version of yjs, but I will be happy to use something easier and more robust than this if yjs core functions allow that.

dmonad:
I like the idea of your approach. If this works for you, you should definitely use it. At some point you could switch to the Yjs based snapshotting approach. 

[...]

PR #225, opened 2020-07-27
restoring document to a specific state using a Snapshot, #159

calibr:
- added public function `createDocFromSnapshot(originDoc, snapshot): Doc`
- added private function `applyDeleteItem(transaction, structs, deleteItem)` to avoid code duplication in places where delete items are needed to be applied to a Doc

### DIFF
--- a/src/utils/DeleteSet.js
+++ b/src/utils/DeleteSet.js
@@ -278,37 +278,13 @@ export const readAndApplyDeleteSet = (decoder, transaction, store) => {
     const state = getState(store, client)
     for (let i = 0; i < numberOfDeletes; i++) {
       const clock = decoder.readDsClock()
-      const clockEnd = clock + decoder.readDsLen()
+      const len = decoder.readDsLen()
+      const clockEnd = clock + len
       if (clock < state) {
         if (state < clockEnd) {
           addToDeleteSet(unappliedDS, client, state, clockEnd - state)
         }
-        let index = findIndexSS(structs, clock)
-        /**
-         * We can ignore the case of GC and Delete structs, because we are going to skip them
-         * @type {Item}
-         */
-        // @ts-ignore
-        let struct = structs[index]
-        // split the first item if necessary
-        if (!struct.deleted && struct.id.clock < clock) {
-          structs.splice(index + 1, 0, splitItem(transaction, struct, clock - struct.id.clock))
-          index++ // increase we now want to use the next struct
-        }
-        while (index < structs.length) {
-          // @ts-ignore
-          struct = structs[index++]
-          if (struct.id.clock < clockEnd) {
-            if (!struct.deleted) {
-              if (clockEnd < struct.id.clock + struct.length) {
-                structs.splice(index, 0, splitItem(transaction, struct, clockEnd - struct.id.clock))
-              }
-              struct.delete(transaction)
-            }
-          } else {
-            break
-          }
-        }
+        applyDeleteItem(transaction, structs, { clock, len })
       } else {
         addToDeleteSet(unappliedDS, client, clock, clockEnd - clock)
       }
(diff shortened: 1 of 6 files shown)


src/utils/DeleteSet.js
```diff
@@ -278,37 +278,13 @@ export const readAndApplyDeleteSet = (decoder, transaction, store) => {
...
-              if (clockEnd < struct.id.clock + struct.length) {
-                structs.splice(index, 0, splitItem(transaction, struct, clockEnd - struct.id.clock))
-              }
-              struct.delete(transaction)
-            }
-          } else {
-            break
-          }
-        }
+        applyDeleteItem(transaction, structs, { clock, len })
```
dmonad, line 287: This may sound silly, but creating a function call for every deleted `Item` really hurts performance.

Yjs is able to handle millions of `Item` integrations because the "hot-path" is inlined. I realize that there is more code that could be inlined. 

Please don't create a separate function call for deleting items.

src/utils/Snapshot.js
```diff
@@ -148,3 +153,124 @@ export const splitSnapshotAffectedStructs = (transaction, snapshot) => {
     meta.add(snapshot)
   }
 }
+
+/**
+ * @param {Doc} originDoc
+ * @param {Snapshot} snapshot
+ * @return {Doc}
+ */
+export const createDocFromSnapshot = (originDoc, snapshot) => {
```
dmonad, line 162: `Item`s depend on each other and must be applied in a very specific order. The function `resumeStructIntegration` handles the correct integration process. If an `Item` depends on another `Item`, the dependency is integrated first.

In multi-user scenarios, the below integration process will likely result in issues. 

Copying `Item`s might be the fastest approach. But I think first we should implement a simple working implementation that just encodes the snapshotted document and then calls `applyUpdate` to make use of the existing struct integration process.

This is how I would lay out the implementation:

* Starting with a snapshot `{ sv, ds }`.
* Split items to ensure that really just the state is copied that we are interested in: `sv.forEach((clock, client) => getItemCleanStart(transaction, ID(client, clock)))` (note: first check if item `ID(client, clock)` exists, otherwise getItemCleanStart might throw an error).
* Create V2 encoder (v2 is probably faster).
* Write all structs from [0 - clock) to encoder.
* Write `ds` to encoder.
* The encoded document is now a document update containing all updates from the state vector
* Apply encoded document to a fresh Y.Doc

tests/snapshot.tests.js
```diff
@@ -0,0 +1,110 @@
+import { createDocFromSnapshot, Doc, snapshot, YMap } from '../src/internals'
```
dmonad, line 1: Thanks for creating tests! These are very valuable!

We should also generate tests that contain multiple users that create changes that depend on each other. E.g. user1 inserts text and syncs with user2. User2 inserts text (the inserted `Item` depends on the content from user1).

calibr: @dmonad thanks, do you mean we need to create a document with these dependent changes and then restore it to a specific state from the past? By generating tests, do you mean that they should be automatically generated somehow(in tests I saw `applyRandomTests` probably used for tests generation)?

src/utils/DeleteSet.js
calibr: Hm, actually, I always thought that modern JS engines do function inlining itselves, so I didn't worry about that too much. With your updated approach using encoding/decoding we won't need `applyDeleteItem`, so I will remove this function.

dmonad: You are right. JavaScript engines do inline code on a predictive basis. In nodejs you can inspect how many functions "can" be inlined and how many functions are actually inlined on execution. It seems there is an inline-budget (because inlining does duplicate code and increases the size of the heap). So inlining functions beforehand, if possible, does improve performance. I just started a few weeks ago to work on inlining the hot-path myself to see some slight improvements of 10-20%. I seriously don't recommend doing that for all projects. But it made sense for Yjs.

tests/snapshot.tests.js
dmonad: >  thanks, do you mean we need to create a document with these dependent changes and then restore it to a specific state from the past? 

Exactly!

> By generating tests, do you mean that they should be automatically generated somehow(in tests I saw applyRandomTests probably used for tests generation)?

You don't have to at first. But it does make sense to do it before we release this feature. A lot of issues can be found with randomized testing. 

Maybe we can adapt `applyRandomTests` to create a snapshot at some point. At the end of the test we restore a document and check if the old values can be recovered.

calibr:
@dmonad hey Kevin, I updated the PR to address your change requests. Now it uses encoding/decoding to restore a snapshot and I added a tests for restoring documents with depended changes

PR #237, opened 2020-09-03
Added draft of INTERNALS.js

josephg:
I wrote up a document describing some of what @dmonad explained to me about how Yjs works. I might be wrong about some details, and missing bits but I don't want to let perfect be the enemy of the good. (Having an incomplete document is better than nothing).

Kevin - please correct / edit any parts of this that I got wrong or things I missed!

### DIFF
--- a/INTERNALS.md
+++ b/INTERNALS.md
@@ -1,0 +1,76 @@
+# Yjs Internals
+
+This document roughly explains how Yjs works internally.
+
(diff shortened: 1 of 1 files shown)


dmonad:
Thanks so much for putting this together!

I extended it with some additional information, but you got everything right. I think I understand what you meant with run length encoding the characters. I removed that part. It might be confusing because RLE encoding also refers to the compression form I use for encoding the integers. Although possible, I don't do any compression on the text representation (well.. except from UTF8 encoding which is a form of compression too). All characters that are ever created are concatenated as a single string (in the V2 encoder) and then written in one flush. The utf8 encoder is faster if it just has to encode a single character. Also it gives the option to encode all characters using some other form of compression (not RLE, but maybe Huffman Coding in the future, as a V3 encoding technique). The V2 encoder is also extensible enough to support/exchange additional encoding techniques.

PR #247, opened 2020-10-10
feat: Add additional options to querySelector

tommoor:
Completes the `TODO` to add filtering by `id` and `attributes`.

~Let me know what you think of this approach, if all is well with you I can apply the same approach to the TODO on `querySelectorAll`  as well – I'm in need of this functionality.~ < Went ahead and did this.


src/types/YXmlFragment.js
```diff
@@ -179,19 +186,60 @@ export class YXmlFragment extends AbstractType {
...
+      filters.tagname = filters.tagname.toUpperCase()
+    }
+
+    const iterator = new YXmlTreeWalker(this, element => {
+      // @ts-ignore
+      if (filters.tagname && element.nodeName && element.nodeName.toUpperCase() === filters.tagname) {
+        return true
+      }
+
+      if (filters.id && element._item && element._item.id === filters.id) {
```
tommoor, line 220: Could be that there is a more direct way of achieving this that I'm missing?

tommoor:
Any word? It'd be good to know if you're considering this for merging or If I should just fork the library

src/types/YXmlFragment.js
dmonad: The lamport timestamps should be completely hidden from the user. There are multiple optimizations in Yjs that will lead to issues when using lamport timestamps in this way. 

If you want to retrieve an element by its unique ID (the lamport timestamp), then you should rather call `getItemCleanStart` on the struct store.  This will allow you to retrieve an element in `O(1)` without querying the XML tree.

dmonad: You currently only do an identity check. You probably meant to compare the ids by value (using `compareIDs(id1, id2)`)..

dmonad (changes requested):
I feel that this implementation of `Filters` does the same as the idea to use `CSS_Selectors`. Unfortunately, I haven't found time to finish up the implementation of querySelector to support the full feature set of `CSS_Selectors`.

But this API might be easier to use in some cases. I dislike the overhead of CSS selectors a bit. In most cases, you simply want to query for some attributes.

I will accept this feature, but the Filters object should be defined as `{ nodeName: string?, attributes: Object<string, any>? }`. I don't want to introduce two similar properties for node-names in YXml. Everything should just be a `nodeName`. My tooling set doesn't use typescript, so I'd rather not use the typescript specific `Record` type. As explained below I don't want to expose the lamport timestamps to the user - they should be kept internal to Yjs and only be used in the form of relative positions.

src/types/YXmlFragment.js
tommoor: Yea, I understand this better now than I did 3 weeks ago. I can remove the `id` querying altogether.

tommoor:
Closing for now, using `xml.createTreeWalker` is the way forward

PR #252, opened 2020-10-27
feat: Add getUsers method to get all users associated with a doc

tommoor:
It's a common usecase to know which users have touched a doc, if using `PermanentUserData` this adds a handy accessor for the information.

I know you're interested in refactoring `PermanentUserData` to a different storage method – the more clear interfaces we can provide now the easier it will make refactoring the internals in the future.

### DIFF
--- a/src/utils/PermanentUserData.js
+++ b/src/utils/PermanentUserData.js
@@ -139,4 +139,21 @@ export class PermanentUserData {
     }
     return null
   }
+
+  /**
+   * @return {Array<string>}
+   */
+  getUsers () {
+    const users = new Map()
+
+    for (const [userDescription] of this.dss.entries()) {
+      users.set(userDescription, true)
+    }
+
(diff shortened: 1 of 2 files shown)


dmonad (changes requested):
I don't use `for .. of` loops because old typescript releases don't convert them according to the js spec. I like to use yjs on stackblitz (and they use this faulty typescript release :disappointed: )

If I remember correctly after calling `setUserMapping` the client automatically creates an entry in `this.clients`. Therefore it should suffice to only use that property, right? It shouldn't be necessary to iterate over `this.dss`

- [ ] Use `forEach`
- [ ] Check if it suffices to only listen to `this.clients`
- [ ] This could be a one-liner: `Array.from(this.clients.keys())`, or `Array.from(new Set([..this.dss.keys(), ..this.clients.keys()]))`

PR #253, opened 2020-10-31
update http links, where possible, to https

lpmi-13:
cattaz.io, unfortunately, is still only available over http, but I've raised an issue in the repo to enable https on github pages, which the site appears to be using.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,12 +48,12 @@ Sponsorship also comes with special perks! [![Become a Sponsor](https://img.shie
 
 ## Who is using Yjs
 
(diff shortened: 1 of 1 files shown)


ISNIT0:
@lpmi-13 You and your bots :)

lpmi-13:
BantzBot said: "A bot wouldn't have had the class to check whether cattaz.io is accessible via https...but I will ask my creator to implement that"

dmonad:
Thanks for the fixes!

PR #268, opened 2020-12-12
Update README.md

jsilvao:
Add a new Yjs user

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ Sponsorship also comes with special perks! [![Become a Sponsor](https://img.shie
   collaboratively organize radio broadcasts.
 * [Cattaz](http://cattaz.io/) A wiki that can run custom applications in the
   wiki pages.
+* [Alldone](https://alldoneapp.com/) A next-gen project management and collaboration platform.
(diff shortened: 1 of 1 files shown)


dmonad:
Hi @jsilvao ,

thanks so much for sharing! Will Yjs be used for the Notes feature?

jsilvao:
Hi @dmonad,

Yes, it will be used for the Notes feature. Actually, the link is pointing to the production deployment. You can check out the Notes feature in the [staging deployment](https://staging.alldone.app). I had to do some mods in the source code of one of your deps (simple-peer) and also mod the signaling mechanism to use some stateful microservices within Google Cloud Functions, but other than that is pure Yjs 😁👍. I'm using it paired with `y-webrtc` and `y-quill`.

dmonad:
Just tested it, really cool! Seems you are using Yjs & y-webrtc in combination with Firestore. Neat idea.

I encountered an error when opening an empty document with two windows. But once I inserted some content and reloaded the windows it worked all fine :+1: 

Thanks again for sharing!

Issue #263, opened 2020-12-02
Differential updates


dmonad:
This is a ticket to track a new feature that will allow users to work with Yjs' document updates without loading the document to memory.

The state of the Yjs document, and also incremental modifications to the document are expressed as  "updates". The idea is now to allow clients to merge updates and compute the differences:

```js
// Instead of applying document updates to a Yjs update, you can simply merge updates:
currentState = Y.mergeUpdate([currentState, updateFromRemotePeer, someOtherUpdate])

// When syncing with another peer: Instead of loading the Yjs document to compute the differences, you can operate directly on the Yjs update:
const diffUpdate = Y.diffUpdate(currentState, remoteStateVector)

// We also need a method to encode the state vector directly from an update:
const stateVector = Y.encodeStateVectorFromUpdate(currentState)
```
The idea of the name `Y.diffUpdate` comes from set theory. I also like the name `Y.sliceUpdate`.

With this, it should be possible to modify the y-websocket server to not load the document to memory at all. Furthermore, we can implement more sophisticated providers that operate on document updates directly instead of the Y.Doc instance.

This is a WIP. I will post updates here.

dmonad:
Short update: I extended the scope of the feature a bit.

Now you are able to merge updates, even if it would not be possible to construct a Yjs document update from it. For example, assume we start with an empty document and then a single client creates `update1`, `update2`, `update3`, and `update4`.

In Yjs, updates from a single user must arrive in creation-order. If we try to apply `update2` before `update1`, then we wait for `update1` to arrive before applying `update2`.

With the new implementation of the feature described above, we can merge updates in any order. I.e.

```js
const update24 = Y.mergeUpdates([update2, update4])
const update13 = Y.mergeUpdates([update1, update3])
const update1234 = Y.mergeUpdates(update24, update13)

equals(update1234, Y.mergeUpdates([update1, update2, update3, update4]))
```

The implementation of `Y.mergeUpdates` is now functional and very well tested on v1 encoding and v2 encoding.

Now I also want to allow gaps in Yjs. So a user should be able to apply `update2` before applying `update1`.

bionicles:
could you please document a usage example for this in the readme, along with an example of the Redis backend? I'd love to play with it!

dmonad:
Hi @bionicles , this will be super interesting for y-redis, but it still uses the old API. I documented the differential updates API with usage examples here: https://docs.yjs.dev/api/document-updates

Issue #265, opened 2020-12-09
Allow transactions to be async


andrew-knack:
**Checklist**

[ ] Are you reporting a bug? Use github issues for bug reports and feature requests. For general questions, please use https://discuss.yjs.dev/
[ ] Try to report your issue in the correct repository. Yjs consists of many modules. When in doubt, report it to https://github.com/yjs/yjs/issues/

**Is your feature request related to a problem? Please describe.**
For multiple changes to considered to be merged as a single change over an arbitrary amount of time. Currently transaction doesn't allow that.

**Describe the solution you'd like**
Allow the [transaction function](https://github.com/yjs/yjs/blob/main/src/utils/Transaction.js#L378-L405) to take in a async function.

**Describe alternatives you've considered**

Not an alternative but I did test that changing the transaction function to async and awaiting the `f` call seemed to work.

**Additional context**
The amount of time that it will take for the transaction is arbitrary based on the user actions so changing the captureTimeout on the `UndoManager` isn't an option.

dmonad:
What is the use-case for this?

andrew-knack:
I need more control over what is considered a single entry in the undo manager. 
In my use case I essentially want to start a transaction on mouse down and end the transaction on mouse up.

dmonad:
Yeah, that makes sense.. But I think that we rather implement this feature in the UndoManager and don't mess with transactions.

Transactions are meant to be atomic changes. If they would be asynchronous, we would need to deal with a lot of new issues (e.g. what happens when two async transactions happen concurrently?).

I'm thinking that we should rather implement something like `undoManager.startCapturing()`. When this method is called, we would capture until `undoManager.stopCapturing()` is called. What do you think? Also, would you be interested in submitting a PR for that?

andrew-knack:
> Transactions are meant to be atomic changes. If they would be asynchronous, we would need to deal with a lot of new issues (e.g. what happens when two async transactions happen concurrently?).

Wouldn't it be the same as transactions are now, the outer captures the inner transaction regardless if it's async or not?

> What do you think? Also, would you be interested in submitting a PR for that?

I've currently got some deadlines I need to meet but during the december break I can give it a go, let's just settle on a design first. 

> I'm thinking that we should rather implement something like undoManager.startCapturing(). When this method is called, we would capture until undoManager.stopCapturing() is called.

There is a part of my use case that I implied but didn't outright say, the transaction also gives control to the user when a change is emitted.

dmonad:
[...]

PR #274, opened 2021-01-29
Differential updates

dmonad:
Implements #263


dmonad:
I wasn't sure how difficult it would be to implement differential updates. 

Some technical details if anyone is interested:

In Yjs we make the assumption that "as long as all updates eventually arrive, all clients are synced". This assumption is really hard to implement because sometimes it is not possible to apply document updates (for example, when a dependency to another update is missing). Usually, these operations are just stored on the StructStore as "pending structs" and are ignored until all dependencies are met.

Now I realized that we might run into troubles when the document is persisted to disk before we find all dependencies for these "pending structs". 

So the new approach is to also encode pending structs as well as all already applied structs. There is a complicated mechanism to make sure that each struct (identified by a lamport timestamp) is applied only once to the document. Filtering duplicates from pending structs is difficult because the compound representation of structs allows Yjs to merge adjacent structs in many cases. So we need to be able to "split" these merged operations whenever necessary.

Also, I needed a method to encrypt efficiently when updates are missing. This is now handled by the `Skip` struct. A Skip tells the decoder that they should just increase the current clock by a certain length (the length of the skip) and then continue. Now the crux: an update that contains a Skip can never be applied because the integration process will fail because some of the encoded structs depend on other structs that do not exist yet. Actually, these structs might be lost forever because the client that created them never comes online again; hence making the document impossible to decode.

So I needed to update the integration process to allow for missing structs. It should just apply as many operations as possible, and put the rest on "pending structs". I hate to work on the integration process `integrateStructs` because it is probably the most difficult function in the codebase. The implementation ended up to be quite beautiful: Whenever the current stack cannot be applied we know that none of the structs can be integrated and we just need to add them to "pending structs". Now we just need a performant memory-saving data store to cache these pending structs. Luckily I already implemented a new struct encoder/decoder: LazyStructDecoder/LazyStructEncoder!

The LazyStructDecoder allows to walk through all structs in an update and extract the content without parsing the complete update (which is often problematic in JavaScript because creating objects expensive).

The LazyStructEncoder allows writing structs one-by-one. The other approach to encode structs is `writeStructs` which requires you to have all structs before you write them. This is necessary because the V2 encoder compresses a lot of metadata. The LazyStructEncoder writes the metadata only when generating the final update message.

With LazyStructDecoder/LazyStructEncoder it is possible to implement an extremely memory-efficient version of `Y.mergeUpdates([update1, update2, ..])`. Basically, we create a LazyStructDecoder for each update and write the struct with the lowest id to the lazyStructEncoder (of course, also filtering out duplicates and merging whenever possible).

Because the lazy decoders are so easy to use, I created a `Y.logUpdate` function (logging the content of an update message) and a `Y.getMetadata` function (logging the starting- and the target-state-vector of an update message). All of this while consuming a constant amount of memory.

The new `Y.diffUpdate(update, sv)` function just iterates through an update (using the lazy decoder) and writes all structs to an encoder that are not filtered by the state-vector. Again, this function uses constant memory.

As I explained in #265, I want to use this implementation to enable the server to sync with clients without parsing the complete document update. The server will be able to sync with clients while using constant memory (just the size of the messages). This will allow y-websocket server to serve many more documents to clients than before.

Another use-case of differential updates is to allow providers to sync thousands of "subdocs" without loading them all to memory. Subdocs are loaded lazily to memory. So often you just want to sync them without loading the document to memory.

I added quite a lot of new tests to make sure that differential updates work as expected. I will give these changes another review and then publish a new release. After that I will work on the y-websocket server improvements.

AdventureBeard:
Thanks for the writeup! This is looking awesome. Also, to clarify, [Shiso](https://shiso.sh/) sponsored this work, as well! 😄

PR #278, opened 2021-02-14
Add slate-yjs to bindings

BitPhinix:
Adds slate-yjs to the bindings inside in the readme.


dmonad:
Sure thing! 

I will also add it to docs.yjs.dev

Issue #287, opened 2021-03-30
Error: "Expect id to be an array" - Intermittent error causing sync to fail across all users in the room


yousourceinc:
**Describe the bug**
I'm receiving an "Expect id to be an array" console error randomly while users are active in a Quill room and it breaks all syncing for all users. 

**Screenshots**
![image](https://user-images.githubusercontent.com/13096930/113016108-bca2f580-91b0-11eb-9183-964869d52893.png)
![image](https://user-images.githubusercontent.com/13096930/113018469-2de3a800-91b3-11eb-973e-e6da4e782dc8.png)
![image](https://user-images.githubusercontent.com/13096930/113018997-b5c9b200-91b3-11eb-9fa6-166d113cdefe.png)

**Environment Information**
Environment Information
Browser: Chrome
Node.js: 12.8.2
Angular: 7
Modules:
"y-array": "10.1.4",
"y-map": "10.1.3",
"y-memory": "8.0.9",
"y-richtext": "9.0.8",
"y-websockets-client": "8.0.16",
"socket.io-client": "2.4.0",
"yjs": "12.3.3"

**Additional context**
The y.connector.isSynced becomes false after the console error. Is there a way to resolve the connection/repair the error somehow? Hope you can point me in the right direction. Thanks!

dmonad:
Hi @yousourceinc I'm extremely overloaded by maintaining too many projects. I can't maintain Yjs v12 that has been deprecated for over a year now. Please upgrade to the current version.

yousourceinc:
hi @dmonad, I understand. Unfortunately, we're pressed for time to fix this since we have active users losing content, and reworking the feature now is not ideal.  

Good thing is, I finally found the cause of the issue in Transaction.js. The implementation during the garbage collection is using a for...in loop which is enumerating object properties instead of iterating items in an array. for...of loop should do the work as per my testing. 

![image](https://user-images.githubusercontent.com/13096930/113033301-72773f80-91c3-11eb-8bfc-93f214c88c3b.png)
![image](https://user-images.githubusercontent.com/13096930/113033465-9f2b5700-91c3-11eb-9057-4ba0a097b05c.png)

I created a PR https://github.com/yjs/yjs/pull/288 for it for the V12 branch. Would it be possible for you to release a new version?

dmonad:
I won't release a new release because it will lead to more trouble for me. As I said, I don't have the time to maintain this for free anymore. 

[...]

PR #288, opened 2021-03-30
V12: Change for...in loop to for...of during GC

yousourceinc:


dmonad:
As I explained in https://github.com/yjs/yjs/issues/287 I don't want to continue maintaining the v12 version anymore. I will only accept security fixes. However, feel free to fork v12.

PR #289, opened 2021-04-01
fix crash when the walker didn't match because n will be null

jhaynie:
When you use a tree walker and it doesn't match any nodes, it will crash. This prevents the crash since null was being check but not for the type before iterating the loop.


dmonad:
Makes sense, thanks! I will publish a new release in a bit.

PR #298, opened 2021-05-05
Expose abstracttype created event

YousefED:
@dmonad Would you be willing to expose this?

For developing libraries similar to MobY, it's necessary to learn about each Y.Map, Y.Array, etc. to be instantiated. Trying to hook into this purely from external code becomes extremely messy, because many of these types are instantiated internally in Yjs. I hope this can provide a clean solution, while we can still keep the other wrappers / patches in a different module

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -25,6 +25,7 @@ export {
   ContentAny,
   ContentString,
   ContentType,
+  typeRefs,
   AbstractType,
   RelativePosition,
   getTypeChildren,
--- a/src/index.js
+++ b/src/index.js
@@ -1,7 +1,7 @@
-
 export {
   Doc,
   Transaction,
+  observeTypeCreated,
   YArray as Array,
   YMap as Map,
   YText as Text,
@@ -25,7 +25,6 @@ export {
   ContentAny,
   ContentString,
   ContentType,
-  typeRefs,
   AbstractType,
   RelativePosition,
   getTypeChildren,
@@ -82,5 +81,5 @@ export {
   encodeRelativePosition,
   decodeRelativePosition,
   diffUpdate,
-  diffUpdateV2
-} from './internals.js'
+  diffUpdateV2,
+} from "./internals.js";
--- a/src/types/AbstractType.js
+++ b/src/types/AbstractType.js
@@ -249,6 +249,21 @@ export const callTypeObservers = (type, transaction, event) => {
 }
 
 /**
+ * Event handlers
+ * @type {EventHandler<AbstractType<any>, undefined>}
+ */
+const typeCreatedEventHandler = createEventHandler();
+
+/**
+ * Observe all events that are created on this type.
+ *
+ * @param {function(AbstractType<any>, undefined):void} f Observer function
+ */
+export function observeTypeCreated (f) {
+  addEventHandlerListener(typeCreatedEventHandler, f)
+}
+
+/**
  * @template EventType
  * Abstract Yjs Type class
  */
@@ -285,6 +300,8 @@ export class AbstractType {
      * @type {null | Array<ArraySearchMarker>}
      */
     this._searchMarker = null
+
+    callEventHandlerListeners(typeCreatedEventHandler, this, undefined);
   }
 
   /**
--- a/src/types/AbstractType.js
+++ b/src/types/AbstractType.js
@@ -249,13 +249,13 @@ export const callTypeObservers = (type, transaction, event) => {
 }
 
 /**
- * Event handlers
+ * Event handler for when new instances of AbstractType are instantiated
  * @type {EventHandler<AbstractType<any>, undefined>}
  */
 const typeCreatedEventHandler = createEventHandler();
 
 /**
- * Observe all events that are created on this type.
+ * Observe instantiation events of new AbstractTypes
  *
  * @param {function(AbstractType<any>, undefined):void} f Observer function
  */


dmonad:
This would fire an event every time a user creates a shared type. Wouldn't it be better to listen to an event that is fired every time a type is added to a specific Yjs document?

YousefED:
This is intentional, although probably your approach could work as well. My approach is a bit easier, because you wouldn't have to call something like registerListeners(Y.Doc) for every document that's created. But both could work.

Maybe it helps if I give you a demo of the purpose when I'm a little bit further :)

dmonad:
Yes, please do! 

I would like to avoid leaking global state like this. There might be an application embedding your library, that doesn't want that all objects are observed.

YousefED:
Hi! It would be nice if we can find a solution for this so `reactive-crdt` doesn't have to rely om my custom branch anymore. How would you go about adding this to Y.Doc?

Btw, the code where I need it is here: https://github.com/YousefED/reactive-crdt/blob/df76836fef31529c1c2ed8187bdce82f3400919f/packages/yjs-reactive-bindings/src/index.ts#L38

If we have support for it, with `yjs-reactive-bindings` we can instantly support Vue-reactive, MobX (it will also replace my previous MobY library), and reactive.

YousefED:
Thanks Johannes!

The issue with only tracking the re-exported yjs, is that yjs also creates
Types internally. For example, if a remote user adds a ymap / yarray /
ytext. These wont be tracked because Yjs would instantiate them using the
internal types.

Good point on hoisting though, thats actually also the reason I reexport
Yjs, so the user doesn’t need to add yjs to their project and can just use
the reactive-crdt version. Probably when mixing different Yjs / reactive
crdt versions it would lead to incompatibilities down the line anyway, so
for now I think its ok to recommend users to use the reexported Yjs.

However, maybe Kevin’s proposed solution would be the best way - I just
dont know how to track all types added to a YDoc.

dmonad:
It would be possible to observe type-creations from the `afterTransaction` event. 

I wonder why you overwrote the Yjs methods? This seems really dangerous to me.  Can't you just use `type.observeDeep(events => ..)`  to listen to all changes that are created by the `type` or any of its children?

In the editor bindings, I try to register as few observers as possible, because registering too many observers often leads to memory issues (because the observers reference objects that shouldn't be in scope anymore). Even nowadays, I still get reports from observers that are not properly unregistered in some cases.

YousefED:
> I wonder why you overwrote the Yjs methods? This seems really dangerous to me. Can't you just use `type.observeDeep(events => ..)` to listen to all changes that are created by the `type` or any of its children?

To bridge with other FRP (functional reactive programming) libraries, we need to know not only what data has changed, but also _which data has been accessed_ (or in other words, needs to be "observered"). For example, to bridge MobX with Yjs, I implement the Atom pattern: https://mobx.js.org/custom-observables.html. This consists of two parts: _reportObserved_ and _reportChanged_.

As an example, let's say I want to make sure the following mobx code automatically responds to changes in the Yjs map:

```
autorun(() => { // autorun is a MobX construct that automatically reruns a piece of code in case observed variables change
console.log(doc.getMap("users").get("peter"));
});

There are two steps involved in this. In the first run, I need to start tracking which variables need to be observed. So the "peter" value in the "users" map needs to be Observed. This is why I patch getMap and Map.get, to be able to call _reportObserved_. Then, I subscribe to Yjs changes and call _reportChanged_ if any of the underlying data has changed.

PS: Yes, purely for reactive-crdt it might be possible to use just observeDeep, as I'm in control of all accessors. However, I also wanted to make it possible to use Yjs in FRP frameworks even if you want to keep using the Yjs API (getMap, getArray) instead of the reactive-crdt abstraction to Javascript constructs (object.property.value instead of nested maps). My original project MobY was such a bridge between the Yjs API and MobX observables (https://github.com/yousefED/moby, note that code will be superseded by the yjs-reactive-bindings subpackage of reactive-crdt as it's pretty much the same). Maybe to clarify:
- yjs-reactive-bindings (reactive-crdt/packages/yjs-reactive-bindings) can bind the Yjs API (doc.getMap, getArray) to FRP libraries (MobX, Vue, or my own [reactive](https://github.com/yousefED/reactive) library)
- reactive-crdt (reactive-crdt/packages/reactive-crdt) builds on yjs-reactive-bindings and yjs to provide a more idiomatic API to Yjs (except for some details it hardly has any code related to FRP in there, it mostly translates YArray to [], and YMap to property.value, using Javascript Proxies)

**Comment by dmonad** [comment]
oh, got it. Thanks for the explanation. Let me think about this a bit.

**Comment by YousefED** [comment]
Thanks @dmonad! This seems to be working, with 2 changes and 1 caveat:
- I had to change to event to `beforeObserverCalls` because I want it to be fired _before_ other observers are triggered by the transaction
- `structs[i].content` can be undefined (in case of GC)

For future reference, the code I have now is:

```
doc.on("beforeObserverCalls", tr => {
    tr.afterState.forEach((clock, client) => {
      const beforeClock = tr.beforeState.get(client) || 0;
      if (beforeClock !== clock) {
        const structs = /** @type {Array<GC|Item>} */ tr.doc.store.clients.get(client);
        const firstChangePos = Y.findIndexSS(structs, beforeClock);
        for (let i = structs.length - 1; i >= firstChangePos; i--) {
          structs[i].content?.getContent().forEach(content => {
            if (content instanceof Y.AbstractType) {
              console.log(content, "is a created type type");
            }
          });
        }
      }
    });
  });
```
(the `structs[i].content?` is the undefined check).

This seems to be working, but I think the only caveat is that it isn't triggered for root types. E.g. when you call `doc.getMap()`, a Y.Map is created, but it doesn't trigger the event. I can probably work around it for my project, but if you have suggestions to catch this scenario as well, that would be great for a complete solution.

**Comment by dmonad** [comment]
I'm glad this works for you as well.

> structs[i].content can be undefined (in case of GC)

Right. Thanks for sharing the code sample. Another edge-case is that structs[i] is already deleted. If a type is created and deleted in the same transaction, then you should just ignore it. Like so:

```js
doc.on("beforeObserverCalls", tr => {
    tr.afterState.forEach((clock, client) => {
      const beforeClock = tr.beforeState.get(client) || 0;
      if (beforeClock !== clock) {
        const structs = /** @type {Array<GC|Item>} */ tr.doc.store.clients.get(client);
        const firstChangePos = Y.findIndexSS(structs, beforeClock);
        for (let i = structs.length - 1; i >= firstChangePos; i--) {
          cost struct = structs[i]
          if (!struct.deleted) {
            // content is not undefined because the type is not marked as deleted
            struct.content.getContent().forEach(content => {
              if (content instanceof Y.AbstractType) {
                console.log(content, "is a created type type");
              }
            });
          }
        }
      }
    });
  });
```

> This seems to be working, but I think the only caveat is that it isn't triggered for root types. E.g. when you call doc.getMap(), a Y.Map is created, but it doesn't trigger the event. I can probably work around it for my project, but if you have suggestions to catch this scenario as well, that would be great for a complete solution.

You could check during `beforeObserverCalls` whether new root-level types were created by iterating through `ydoc.share`. E.g. 


```js
doc.on("beforeObserverCalls", tr => {
   ydoc.share.forEach(type => {
      // the explicit check is necessary because we sometimes initialize "anonymous" types that the user can't (and shouldn't) access. 
      if (type.constructor !== Y.AbstractType) {
         if (!allMyTypes.has(type)) {
           allMyTypes.add(type)
         }
      }
   })
})
```

YousefED:
@dmonad I encountered another edge case where initial structs are not detected. E.g., let's say there's a Y.Text stored in a root level map, we should also "hook" those objects initially.

I have solved this by initially iterating through `doc.store.clients` and all structs stored in there for every client. Does that make sense, or should I filter by clientId and / or clock?

PR #309, opened 2021-06-08
add `clear` to YMap

whawker:
Remove all elements from a YMap 

Uses `delete` via `forEach` to create an event history (is this the correct approach?)


dmonad:
Perfect, thanks for the clean PR with tests!

PR #318, opened 2021-07-25
Add valtio-yjs binding in README

dai-shi:
Hi, I'm not sure if you accept adding a new binding which is in alpha.
It's not tied to an app, but it's a two-way binding to a proxy state.


dmonad:
Sweet, this library looks fun! Thanks for creating it and adding it to the list :heart: 

I should probably add another section to the README for non-editor bindings. For now, this is fine.

PR #323, opened 2021-08-09
Update an algorithm example

vivaxy:
Hi @dmonad.

The explanation on 'users inserting elements in sequence' seems not right. The outcome(`[{id: {client, clock: 0}, content: 'ab'}]`) seems not corresponding to the input(`array.insert(0, ['a']), array.insert(0, ['b']);`).

It should be inserting text I think. And the insertion sequence should be `0` and `1`.


dmonad:
Right, thank you for noticing!

PR #326, opened 2021-08-22
[tests] Encode/decode doc with attribute changes

raedle:
Encode a document with a text and decode it into a new document. Then, test if the same change to both documents results in the same text deltas.

This PR adds a test, which currently fails. It is expected that both documents would be the same

```
[
    { insert: '\n', attributes: { url: 'http://example.com' } },
    { insert: '\n', attributes: { url: 'http://docs.yjs.dev' } },
    { insert: '\n', attributes: { url: 'http://example.com' } }
]
```

### DIFF
--- a/tests/encoding.tests.js
+++ b/tests/encoding.tests.js
@@ -106,3 +106,28 @@ export const testDiffStateVectorOfUpdateIgnoresSkips = tc => {
   t.assert(state.get(ydoc.clientID) === 1)
   t.assert(state.size === 1)
 }
+
+/**
+ * Encode a document with a text and decode it into a new document. Then, test if
+ * the same change to both documents results in the same text deltas.
+ *
+ * @param {t.TestCase} tc
+ */
+export const testEncodeDecode = async tc => {
+  const ydoc1 = new Doc()
+  const ydoc2 = new Doc()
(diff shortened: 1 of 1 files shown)


raedle:
@dmonad The idea for this test was to validate the following:

* Client 1 creates a Y.Text and makes changes
* Client 1 syncs with client 2
* Both "clients go offline" and make the same edits

**Shouldn't the same edit (even if offline) result in the same delta for text1 and text2?**

The delta for `text1` is what I would expect.

The delta for `text2` has lost the attribute for the third line break `\n`.

The delta for text1 is:

```
[
  {
    "insert": "\n",
    "attributes": {
      "url": "http://example.com"
    }
  },
  {
    "insert": "\n",
    "attributes": {
      "url": "http://docs.yjs.dev"
    }
  },
  {
    "insert": "\n",
    "attributes": {
      "url": "http://example.com"
    }
  }
] 
```

The delta for text2 is:

```
[
  {
    "insert": "\n",
    "attributes": {
      "url": "http://example.com"
    }
  },
  {
    "insert": "\n",
    "attributes": {
      "url": "http://docs.yjs.dev"
    }
  },
  {
    "insert": "\n"
  }
] 
```

PR #329, opened 2021-09-13
Add Vertex Software to 'Who is using Yjs'

therockstorm:


dmonad:
Hi @therockstorm,

this is awesome, thank you so much for sharing!

The "In the wild" section is meant to link to production applications that use Yjs. People tend to be more willing to use Yjs when they see that Yjs is trusted by companies. I think your demo should be added to https://github.com/yjs/yjs-demos/. Ideally, you add an entry to the top of the list that links to your demo. 

However, I see that your product also supports collaboration. Please feel free to update this PR to link to https://vertexvis.com/ if it uses Yjs in production.

therockstorm:
It doesn't yet, but it may in the near future! Thanks for taking a look, I'll get it added to yjs-demos.

PR #359, opened 2021-12-09
Add SyncedStore to readme

YousefED:
Not sure what the preferred way to share it would be, for now I added it to the Bindings :)


dmonad:
Thanks! This is fine. I should probably introduce a section for non-editor binding at some point.

PR #368, opened 2022-01-02
Undo event

nc:
Includes doodlewind/yjs#undo_patch

Only emit `stack-item-added` when an item is actually added, vs coalesced with a previous one if `captureTimeout is > 0`

Used to implement an undo manager that is comprised of several Y.UndoManagers. Since this changes the current behavior, unclear if it's worth merging in.


dmonad:
Thanks for this. Makes sense. I published an alternative fix, but carried over your changes. 

I will publish a new release in a bit.

PR #373, opened 2022-01-09
make size() public

ffd8:
perhaps it was a typo? this should allow `size()` of map to be public.


ffd8:
No need to change anything here.. instead will add `.size` as a note in the API docs as [suggested here](https://discuss.yjs.dev/t/getting-y-map-size/926).

PR #386, opened 2022-01-31
Avoid copying attribute map when deleting

### LINKED ISSUE #305: missing attribute when deleting the whole content using applyDelta
**Describe the bug**

When the whole formatted content gets deleted from YText using applyDelta and then new content gets inserted in the same call with the same attribute, this attribute will be missed from the resulting content.

**To Reproduce**

```
const yjs = require('yjs')

const doc = new yjs.Doc()
const text = doc.get('text', yjs.Text)
const str = 'hello world'

text.applyDelta([{
    insert: str,
    attributes: {
      bold: true
    }
}])

text.applyDelta([
  { delete: str.length },
  {
    insert: 'hello',
    attributes: {
      bold: true
    }
  }
])

console.log(text.toDelta())
```

This code prints:
```
[ { insert: 'hello' } ]
```

**Expected behavior**

Result of the code above should be:
```
[ { insert: 'hello', attributes: {bold: true} } ]
```

**Environment Information**

Yjs 13.5.8

**Additional context**

On the line https://github.com/yjs/yjs/blob/main/src/types/YText.js#L469
if i replace `map.copy(currPos.currentAttributes)` to just `currPos.currentAttributes` it seems to be working fine, but I am hesitating because it may break something else.

ja2nicholl:
Calling cleanupFormattingGap should not make a copy of the
attributes because it needs to be able to update them.

Fixes #305 with the suggestion from @calibr.

### DIFF
--- a/src/types/YText.js
+++ b/src/types/YText.js
@@ -465,7 +465,7 @@ const deleteText = (transaction, currPos, length) => {
     currPos.forward()
   }
   if (start) {
-    cleanupFormattingGap(transaction, start, currPos.right, startAttrs, map.copy(currPos.currentAttributes))
+    cleanupFormattingGap(transaction, start, currPos.right, startAttrs, currPos.currentAttributes)
   }
   const parent = /** @type {AbstractType<any>} */ (/** @type {Item} */ (currPos.left || currPos.right).parent)
   if (parent._searchMarker) {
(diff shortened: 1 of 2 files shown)


dmonad:
Thanks for this! 

We mustn't modify the `currentAttributes` unless we walk the item position. Hence we needed to copy it. However, I fixed the edge case using your contribution. I will publish the fix in a bit.

PR #397, opened 2022-03-07
add immer-yjs to Bindings

sep2:
This PR adds [immer-yjs](https://github.com/sep2/immer-yjs) to Bindings section in readme.md.

`immer-yjs` allows manipulating `y.js` data types with the api provided by `immer`.

* Two-way binding between y.js and plain (nested) json object/array.
* Efficient snapshot update with structural sharing, same as `immer`.
* Updates to `y.js` are explicitly batched in transaction, you control the transaction boundary.
* Always opt-in, non-intrusive by nature (the snapshot is just a plain object after all).
* The snapshot shape & y.js binding aims to be fully customizable.
* Typescript all the way (pure js is also supported).
* Code is simple and small, no magic hidden behind, no vendor-locking.


dmonad:
Of course, thank you!

PR #401, opened 2022-03-13
Add Living Spec to readme

dylans:
We've sponsored via GitHub and OpenCollective for a while and have made a few contributions. We'd like to be added to the list of users if appropriate. Thanks.


dmonad:
Right, thank you so much!

PR #418, opened 2022-04-12
Use correct namespace when generating dom nodes for SVG content

aartrost:
Hey Kevin,

Please let me know if this is equally out of scope as https://github.com/yjs/yjs/pull/341 and we'll retract. However we've found it harder to externally override this method because it's dependency on `typeListForEach` which is not public.

Without this patch `toDOM()` will output SVG content that's not renderable by the browser.

### DIFF
--- a/src/types/YXmlElement.js
+++ b/src/types/YXmlElement.js
@@ -201,7 +201,20 @@ export class YXmlElement extends YXmlFragment {
    * @public
    */
   toDOM (_document = document, hooks = {}, binding) {
-    const dom = _document.createElement(this.nodeName)
+    let dom;
+    switch(this.nodeName){
+      case "polyline":
+      case "path":
+      case "rect":
+      case "circle":
+      case "ellipse":
+      case "line":
+      case "svg":
+        dom = _document.createElementNS('http://www.w3.org/2000/svg', this.nodeName);
+        break;
+      default:
+        dom = _document.createElement(this.nodeName);
+    }
     const attrs = this.getAttributes()
     for (const key in attrs) {
       dom.setAttribute(key, attrs[key])


dmonad:
Hi @aartrost,

I want to remove the `toDom` from Yjs in the next release as I feel this can be easily implemented using a custom function.

Instead of `typeListForEach` you can simply use `type.forEach(child  => ..)`.

I'm hesitant to introduce this fix because it is a breaking change. I suggest that you try to move away from the native `toDom` method and let me know if there are any downsides. If there are good arguments for keeping this method, then I'll merge this fix into the next major release (in a couple of months or maybe earlier if you need this fix).

dmonad:
Hi @aartrost,

I want to remove the `toDom` from Yjs in the next release as I feel this can be easily implemented using a custom function.

Instead of `typeListForEach` you can simply use `type.forEach(child  => ..)`.

I'm hesitant to introduce this fix because it is a breaking change. I suggest that you try to move away from the native `toDom` method and let me know if there are any downsides. If there are good arguments for keeping this method, then I'll merge this fix into the next major release (in a couple of months or maybe earlier if you need this fix).

aartrost:
That makes total sense. 

We are still working on a stable version of https://github.com/yjs/y-dom. We're getting pretty close to a place where we can push a stable-ish beta version of our improvements.

If you intend to remove `.toDOM` we can re-implement it in the y-dom binding ( would `Y.XmlElement.prototype.toDOM = () => {...}` be your preferred way of doing that? ). Perhaps we can additionally implement a `Y.XmlElement.prototype.toHTMLString` as a HTML-compliant version of `.toString`.

Let me know what you think

dmonad:
> If you intend to remove .toDOM we can re-implement it in the y-dom binding ( would Y.XmlElement.prototype.toDOM = () => {...} be your preferred way of doing that? ).

That is certainly possible. However, personally, I'd avoid manipulating a prototype that I don't "own". There is always the possibility that y-dom is included several times in different versions. You can guard against these problems by using a simple function to transform any Y.Xml to DOM.

> Perhaps we can additionally implement a Y.XmlElement.prototype.toHTMLString as a HTML-compliant version of .toString.

That makes total sense!

> Edit: I just attempted to implement the above but the YXmlElement type does not have a forEach method. I'm using Y.Array.prototype.forEach.call(this, yxml => {...}) where this is the YXmlElement as a stop-gap for now but it might be nice to have native support for that on the type itself.

Right! Nice workaround though. I just opened a ticket for that #421. I'll fix it in the next release (this week).

dmonad:
> That is certainly possible. However, personally, I'd avoid manipulating a prototype that I don't "own". There is always the possibility that y-dom is included several times in different versions. You can guard against these problems by using a simple function to transform any Y.Xml to DOM.

That requires you to do constructor checks: e.g.

 ```js
switch (yxml.constructor) {
  case Y.XmlElement: {
  },
  case Y.XmlFragment: {
  },
  .. 
}
```

It is safe to assume that you only have one Yjs package included. You actually get a warning if you include Yjs in different versions several times. Constructor checks are, unfortunately, necessary to make other bindings work as well. There is no good workaround for this problem.

aartrost:
Right, extending the prototype was not a good idea. 

I've re-implemented all this as a standalone `convertTypeToDom` function with support for `Y.XmlElement`, `Y.XmlFragment`, `Y.XmlText` and `Y.XmlHook`. Works like a charm and our project is not depending on any `.toDom` calls anymore. A snippet of that function in case it's useful to anybody else: https://gist.github.com/aartrost/e122cd58397bf5abe52ac2c96960c9d5.

I've added a comment to https://github.com/yjs/yjs/issues/421 as we're also in need to of a `forEach` on `Y.XmlFragment`. Thanks for all your help so far!

PR #424, opened 2022-05-17
Add support for esm in node.js

KentoMoriwaki:
I'm trying to yjs with TypeScript's experimental support for ECMAScript modules in node.js.
https://www.typescriptlang.org/docs/handbook/esm-node.html

To import the types correctly, we need to add `exports.types` field.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
   },
   "exports": {
     ".": {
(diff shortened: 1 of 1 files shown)


dmonad:
Got it, so they deprecated the old root-level `types` field in package.json.

Issue #431, opened 2022-06-01
stopCapturing does is not acknowledge when captureTimeout is greater than the current unix timestamp


PatrickShaw:
**Checklist**

* [x] Are you reporting a bug? Use github issues for bug reports and feature requests. For general questions, please use https://discuss.yjs.dev/
* [x] Try to report your issue in the correct repository. Yjs consists of many modules. When in doubt, report it to https://github.com/yjs/yjs/issues/

**Describe the bug**
When `captureTimeout` is larger than the unix timestamp, `.stopCapturing()` is no longer able separates items in the stack.

**To Reproduce**
1. Create a `Y.UndoManager` with any shared type instance attached to it (E.g. `Y.Array`). Set the `captureTimeout` to anything larger than unix timestamp. `Number.POSITIVE_INFINITY` or `Number.MAX_VALUE` will work.
3. Change some data in the shared type instance.
4. Call `.stopCapturing()` undo manager
5. Change some data in the shared type instance.
6. Call `.undo()` on the undo manager
7. At this point, you'll notice that `.stopCapturing` was ignored and that both step 3 and 5 will be undone.

**Expected behavior**
Only step 5 should have been undone

**Screenshots**
-

**Environment Information**
The problem is env agnostic. Node + Chrome on MacOS is what I used.

dmonad:
I suggest that you set captureTimeout to a few weeks. I'll think of a better solution when I'm back from vacation.

PR #432, opened 2022-06-01
Fix YJS UndoManager .stopCapturing being acknowledged when captureTimeout above unix timestamp

PatrickShaw:
This PR adds tests to replicate #431 and then changes the `Y.UndoManager` to abide by those tests.


src/utils/UndoManager.js
```diff
@@ -316,7 +316,7 @@ export class UndoManager extends Observable {
    *
    */
   stopCapturing () {
-    this.lastChange = 0
+    this.lastChange = Number.NEGATIVE_INFINITY;
```
PatrickShaw, line 319: There's a line above in this file that says `now - this.lastChange < captureTimeout`.

Setting `lastChange` to `0` will only work if the `captureTimeout` make `now - this.lastChange` equal to `now` which is the unix timestamp. I hope that's enough to show how #431 arises.

tests/undo-redo.tests.js
```diff
@@ -588,3 +588,59 @@ export const testBehaviorOfIgnoreremotemapchangesProperty = tc => {
...
+ * Ensure that we never stop capturing due to a timeout when the timeout is infinite
+ *
+ * @see https://github.com/yjs/yjs/issues/431
+ * @param {t.TestCase} tc
+ */
+ export const testInfiniteCaptureTimeoutNeverCapturesAutomatically = tc => {
+  const { array0 } = init(tc, { users: 3 })
+  const undoManager = new Y.UndoManager(array0, { captureTimeout: Number.POSITIVE_INFINITY })
+
+  array0.push([1, 2, 3]);
```
PatrickShaw, line 602: Would be nice in the future if the timeout logic were decoupled from the undo manager itself so we could test the behaviour in isolation, without any extra logic related to the shared types getting in the way.

dmonad:
Thank you, this looks pretty good. I'll probably merge it in a few weeks when I'm back from vacation.

src/utils/UndoManager.js
HondryTravis: god pls

HondryTravis: this pr now is not havn't process?

dmonad:
@PatrickShaw Thanks for this <3 Sorry it took so long to merge.. I ported your test and fixed the problem in a different PR. I adapted the solution a bit and made `lastChange=0` a special case. Comparing infinite values (`lastChange=-∞` & `captureTimeout=∞`) seems dangerous and might not always return the expected result.

PR #435, opened 2022-06-14
export UpdateEncoderV1

yogas:
Hi @dmonad

In Nimbus we are using `UpdateEncoderV1` but in new release of `yjs` its gone from exports. 
May we return it?

Thank you


dmonad:
Sure! Thanks for the PR. Going to release this in a bit.

yogas:
Hi @dmonad , Thank you!

PR #437, opened 2022-06-16
Improve double-import-message

mrdrogdrog:
This PR improves the double import error message.


dmonad:
I kept the message intentionally small. Otherwise, people won't read it. I already changed it to be a proper error message (in red!) and made the warning quite serious. If people ignore this message, it's their own fault.

I don't want to link to source-code in github. However, I fixed the spelling mistake and created a ticket for the issue https://github.com/yjs/yjs/issues/438

mrdrogdrog:
That's also fine by me ☺️

PR #439, opened 2022-06-23
Make encodeStateAsUpdate deterministic

adamchel-synth:
This PR modifies the encoding of the delete set in `encodeStateAsUpdate` so that the entries are sorted in descending client ID order. 

This makes it possible to perform hash or checksum on the encoded state of a document and get a consistent result across different peers if the states are equivalent.


dmonad:
Thanks @adamchel-synth, just made a release!

adamchel-synth:
Amazing, thank you!

Issue #441, opened 2022-06-30
Missing export for encodeStateVectorV2


reijne:
**Checklist**

* [x] Are you reporting a bug? Use github issues for bug reports and feature requests. For general questions, please use https://discuss.yjs.dev/
* [x] Try to report your issue in the correct repository. Yjs consists of many modules. When in doubt, report it to https://github.com/yjs/yjs/issues/

**Describe the bug**
The function `encodeStateVectorV2` is defined in src/utils/encoding.d.ts, however, it is not exported in dist/src.
Therefore, when importing from 'yjs', the function does not exist.

**To Reproduce**
1. `import * as Y from 'yjs'`
2. `Y.encodeStateVectorV2(...)`
3. Error: property does not exist

**Expected behavior**
Like the other V2 equivalent functions, I expect the property to be exported in dist/src, allowing usage when importing yjs. For example `encodeSnapshotV2`

**Environment Information**
 - Browser independent. Used: Chrome
 - "yjs": "^13.5.24"

Hideman85:
Same issue here

chris-top-legal:
Same issue here

dmonad:
The internal function Y.encodeStateVectorV1 is currently the same as encodeStateVectorV2. This might change in the future without breaking the exposed API. encodeStateVectorV2 is strictly for internal usage (for consistency). You should use Y.encodeStateVector without suffix.

PR #442, opened 2022-06-30
Add `encodeStateVectorV2` to list of exports

reijne:
As mentioned in bug report 441: https://github.com/yjs/yjs/issues/441

The export for encodeStateVectorV2 is missing.
This pr adds it to the list of exported functions in index.js

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -59,6 +59,7 @@ export {
   encodeStateAsUpdate,
   encodeStateAsUpdateV2,
   encodeStateVector,
+  encodeStateVectorV2,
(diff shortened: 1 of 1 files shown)


tuoxiansp:
also missed `decodeStateVectorV2`

dmonad:
The internal function `Y.encodeStateVectorV1` is currently the same as `encodeStateVectorV2`. This might change in the future without breaking the exposed API. `encodeStateVectorV2` is strictly for internal usage (for consistency). You should use `Y.encodeStateVector` without suffix.

PR #449, opened 2022-08-02
add option to UndoManager to control the behavior of destroy func

LeeSanity:
as [discussed](https://github.com/yjs/y-prosemirror/issues/114#issuecomment-1188911948) before, this option has solved my problem! thanks a lot.

### DIFF
--- a/src/utils/UndoManager.js
+++ b/src/utils/UndoManager.js
@@ -139,6 +139,9 @@ const popStackItem = (undoManager, stack, eventType) => {
  * undo/redo scope.
  * @property {Set<any>} [UndoManagerOptions.trackedOrigins=new Set([null])]
  * @property {boolean} [ignoreRemoteMapChanges] Experimental. By default, the UndoManager will never overwrite remote changes. Enable this property to enable overwriting remote changes on key-value changes (Y.Map, properties on Y.Xml, etc..).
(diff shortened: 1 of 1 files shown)


src/utils/UndoManager.js
```diff
@@ -372,6 +377,8 @@ export class UndoManager extends Observable {
   }
 
   destroy () {
+    if (!this.shouldDestoryUndoManager) return
+
     this.trackedOrigins.delete(this)
     this.doc.off('afterTransaction', this.afterTransactionHandler)
```
dmonad, line 383: The following code should always be executed as it is specific to the binding. 

```js
    this.trackedOrigins.delete(this)
    this.doc.off('afterTransaction', this.afterTransactionHandler)
```

dmonad (changes requested):
Thanks for this!

This looks good. I requested a single change. Can you please implement it and check if it still works for you?

src/utils/UndoManager.js
LeeSanity: Sorry, this won't work for me, because in my case, prosemirror editors are nested in the same Y.Doc with other componenets, if afterTransaction event is unregisterred, then this undoManager will stop working for other components of this Y.Doc. It seems that there only one way left for me, leave the destroy controll to me.

LeeSanity: For now, I must overwrite prosemirror yUndoPlugin's view function to disable destroy calls.

PR #453, opened 2022-08-10
Allow updating captureTimeout on UndoManager instances

aartrost:
UndoManager is currently only using the initial `captureTimeout` passed to the constructor. When moving this into `this.captureTimeout` we can now update the timeout after init. 

For example: we use `yUndoManager.captureTimeout = Number.MAX_VALUE` to temporarily pause stack item reporting for long user interactions.

### DIFF
--- a/src/utils/UndoManager.js
+++ b/src/utils/UndoManager.js
@@ -192,6 +192,7 @@ export class UndoManager extends Observable {
     this.doc = doc
     this.lastChange = 0
     this.ignoreRemoteMapChanges = ignoreRemoteMapChanges
+    this.captureTimeout = captureTimeout
     /**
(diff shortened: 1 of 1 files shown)


Wikigoagec:
> UndoManager is currently only using the initial `captureTimeout` passed to the constructor. When moving this into `this.captureTimeout` we can now update the timeout after init. 
> 
> For example: we use `yUndoManager.captureTimeout = Number.MAX_VALUE` to temporarily pause stack item reporting for long user interactions. 

tail -n kafka/data/file_consumer/test.sink.txt

dmonad:
Makes sense, thanks!

Will get it into the next release.

PR #462, opened 2022-09-09
Export missing `encodeStateVectorV2` method

vkartaviy:


dmonad:
There is no difference in the encoding of state vectors. You can simply use `encodeStateVector`. I don't want to export the v2 version to keep the namespace free for future usage.

PR #465, opened 2022-09-19
Remove unused return in `forEach` of `YMap`

neo:
If the idea is to keep the API as close to the JS Map as possible, maybe we should consider not returning here.

Ref: https://github.com/microsoft/TypeScript/blob/v4.8.3/lib/lib.es2015.collection.d.ts#L28-L31


dmonad:
Yeah, absolutely. Thank you!

PR #471, opened 2022-10-05
Update outdated LICENSE year

Serjobas:


dmonad:
I'm unsure if it was necessary, but I updated the license year in another commit.

PR #482, opened 2022-11-05
Add missing word

joakim:
Added missing "are" (3rd word) and formatted to 80 cols.

Hope I can offer more substantial contributions down the line :)


dmonad:
Oh, thanks for the fix <3

PR #501, opened 2023-02-05
Don't update the current attributes in case of delete

nimishclickup:


dmonad:
Thanks for the PR! I fixed the issue in another way and made a release. But this helped to narrow down the issue more quickly!

PR #505, opened 2023-02-23
fix: deleteText currPos left and right both null

rcbevans:
<sub><a href="https://huly.app/guest/w-githubdmonad-yjs-660d5772-70f06a0a22-d28cf5?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NjBkNjNiODhjN2YzNTBiMTEyMTNlZjUiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctZ2l0aHViZG1vbmFkLXlqcy02NjBkNTc3Mi03MGYwNmEwYTIyLWQyOGNmNSIsInByb2R1Y3RJZCI6IiJ9.pPOVtq940DTKYUP_VHZ9GlanL74sH6FaN7y0jzV8hA0">Huly&reg;: <b>YJS-431</b></a></sub>

### DIFF
--- a/src/types/YText.js
+++ b/src/types/YText.js
@@ -498,8 +498,8 @@ const deleteText = (transaction, currPos, length) => {
   if (start) {
     cleanupFormattingGap(transaction, start, currPos.right, startAttrs, currPos.currentAttributes)
   }
-  const parent = /** @type {AbstractType<any>} */ (/** @type {Item} */ (currPos.left || currPos.right).parent)
-  if (parent._searchMarker) {
+  const parent = /** @type {AbstractType<any>} */ (/** @type {Item} */ (currPos.left || currPos.right)?.parent)
+  if (parent?._searchMarker) {
     updateMarkerChanges(parent._searchMarker, currPos.index, -startLength + length)
   }
   return currPos


YuliaTsareva:
Any chance this PR will be merged? We also encounter issues with these lines of code.

dmonad:
I know other projects accept "quick fixes" that simply fix an issue, so we can move on for another week. But these kinds of fixes often cause more harm because they don't fix the underlying issue. I want to understand what's wrong first.

This is why I require a proper bug report before we can implement a fix. How can I reproduce this issue? 
At least I need to know which Yjs method is called when this issue is called. An error stack would be nice and easy to produce.

This is not the right fix for the underlying issue, I'm sorry. It shouldn't be possible to create `ItemTextListPosition` in empty documents (`left` AND `right` is null). I don't understand how this error can occur. If you can create positions in empty documents with the Yjs API, then we are prone to other issues as well. I'd like to prevent creating positions in empty documents.

PR #515, opened 2023-03-28
Restore deleted entries in a map by the undomanager

### LINKED ISSUE #500: unexpected undo result on YMap while delete entries

**Checklist**

* [ ] Try to report your issue in the correct repository. Yjs consists of many modules. When in doubt, report it to https://github.com/yjs/yjs/issues/

**Describe the bug**
see reproduce
**To Reproduce**
```ts

const testUndoMap = () => {
  const doc = new Y.Doc()
  const yMap = doc.getMap('map')
  const undoManager = new Y.UndoManager(yMap, { captureTimeout: 0 })
  doc.transact(() => {
    yMap.set('a', 'a')
  })
  doc.transact(() => {
    yMap.delete('a')
    yMap.set('b', 'a')
  })

  console.assert(yMap.get('b') === 'a', '.b should be a ', yMap.get('b'), yMap.get('a')) // ok

  doc.transact(() => {
    yMap.set('a', 'a')
    yMap.delete('b')
  })

  console.assert(yMap.get('a') === 'a', '.a should be a ', yMap.get('b'), yMap.get('a')) // ok

  undoManager.undo()
  console.assert(yMap.get('b') === 'a', yMap.get('b'), yMap.get('a')) // ok
  console.assert(yMap.get('a') === undefined, yMap.get('b'), yMap.get('a')) // ok

  undoManager.undo()
  console.assert(yMap.get('a') === 'a', '.a should be a ', yMap.get('b'), yMap.get('a')) // Assertion failed: .a = a  undefined undefined
}

testUndoMap()
```

**Expected behavior**
finally `yMap.get('a') === 'a'`

**Environment Information**
Chrome 109.0.5414.87
Yjs v13.5.44

dhenneke:
Fixes #500

This issue occurred in our data structure where we store a "lock" of an object. This is can be enabled/disabled by the user but a repeated toggle couldn't be undo'd. The test still works in version 13.5.24 and was broken since https://github.com/yjs/yjs/commit/4cfa49d601743e4366109b6e800125322b5c147a.

I find it suspicious that `item.right` is checked for existence but then nothing is done with it. All the other tests still work, so I hope this doesn't introduce new bugs.

### DIFF
--- a/src/structs/Item.js
+++ b/src/structs/Item.js
@@ -200,7 +200,7 @@ export const redoItem = (transaction, item, redoitems, itemsToDelete, ignoreRemo
   } else {
     right = null
     if (item.right && !ignoreRemoteMapChanges) {
-      left = item
+      left = item.right
       // Iterate right while right is in itemsToDelete
       // If it is intended to delete right while item is redone, we can expect that item should replace right.
       while (left !== null && left.right !== null && isDeleted(itemsToDelete, left.right.id)) {
--- a/tests/undo-redo.tests.js
+++ b/tests/undo-redo.tests.js
@@ -647,7 +647,7 @@ export const testSpecialDeletionCase = tc => {
 
 /**
  * Deleted entries in a map should be restored on undo.
- * 
+ *
  * @see https://github.com/yjs/yjs/issues/500
  * @param {t.TestCase} tc
  */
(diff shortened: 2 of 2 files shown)


dmonad:
Thanks @dhenneke ! I cherry-picked the test case from this PR. The condition `item.right !== null` is for checking whether concurrent changes might block undoing the current item (collab UndoManager is just ridiculously complex). So the actual fix was a bit more involved.

dhenneke:
Thanks 🎉 I already expected that it would be a bit more complex than I thought. But I hope that I could help by at least pointing out where the assumed bug was.

dmonad:
Yep, that was very helpful! PRs for test-cases are always appreciated and help me debugging!

PR #523, opened 2023-04-13
Change cleanupTransactions to avoid call stack overflow

### LINKED ISSUE #522: `cleanupTransactions` causes "maximum call stack size exceeded" (+fix PR)

**Checklist**

**Describe the bug**
The `cleanupTransactions()` function in `yjs/yjs` (lines  [244-374 of `src/utils/Transaction.js`](https://github.com/yjs/yjs/blob/main/src/utils/Transaction.js#L244-L374)) is a recursive function that will overflow the call stack in certain browsers when there are a large number of transactions needed to load a document.  This results in the error shown in the image below.

for-PR-submissions

The image is from Firefox; in Chrome the error message is slightly different (`maximum call stack size exceeded`), but the root problem is the same:  excessive recursion overflowing the call stack.

This only happens in Chrome and Firefox; it doesn't happen in Safari.  (The reason why is that Chrome [does not implement tail call optimization](https://chromestatus.com/feature/5516876633341952) whereas Safari does.  Hat tip to my colleague @joshlabau for pointing this out to me.)

This problem has also been reported recently [on StackOverflow in this post](https://stackoverflow.com/questions/75931668/data-persistence-due-to-rangeerror-maximum-call-stack-size-exceeded-y-js).

**To Reproduce**
1. (In the following steps use Chrome as your browser.)
2. I assume you have a browser-based instance of a Yjs + Quill single page app available already somewhere.  If you don't, I can stand up a demo server that illustrates this problem and share it privately with you so you can see without doing a lot of work.  You can contact me at fsctl99@gmail.com.
3. In the following steps, you are going to create a new Y.Doc and then do a bunch of insertions into it.  Either use a backend that persists the Y.Doc even when no one is connected, or keep a browser open with that Y.Doc loaded.  (The point is just that you need to keep the Y.Doc alive during all the remaining steps.)
4. Script a headless browser (I'm using Node `puppeteer`) to  make a large number of short insertions into a Y.Doc, one at a time, reloading the page in between each insertion.  About 10,000 single-sentence paragraph insertions should be more than adequate.
5. In another Chrome tab or another instance of Chrome on a different machine, try to load the same page in order to cause a fresh load of the Y.Doc.  The error occurs on document load.
6. Open the browser developer console and see the error illustrated above.

**Expected behavior**
The call stack should not overflow.  There should be no error at all when opening a large Y.Doc like this.

**Screenshots**
A screenshot taken from Firefox is included above.  A comparable screenshot could be produced with Chrome, but I don't have it on hand.

Also, see the [same bug described by someone on Stack Overflow](https://stackoverflow.com/questions/75931668/data-persistence-due-to-rangeerror-maximum-call-stack-size-exceeded-y-js) as mentioned above.

**Environment Information**
 - macOS Ventura, on a MacBook Pro
 - Browser: latest version of Chrome or Firefox (as stated above)
 - Node v18.16.0 (the node version is not important here)
 - `npm ls yjs`:

    ```
    yjs-quill@1.0.0 /code/yjs-quill
    ├─┬ y-quill@0.1.5
    │ └── yjs@13.5.52 deduped
    ├─┬ y-websocket@1.5.0
    │ ├─┬ y-leveldb@0.1.2
    │ │ └── yjs@13.5.52 deduped
    │ └── yjs@13.5.52 deduped
    └── yjs@13.5.52
    ```

**Additional context**
 - The fix is trivial and I've already implemented it in my fork of `yjs/yjs`.  EDIT:  [here is my simple PR #523](https://github.com/yjs/yjs/pull/523) that fixes this issue.
 - The fix is to change `cleanupTransactions()` from a recursive function to an iterative function.
 - I wanted to open this discussion issue  in addition to the PR because I thought there might be some other larger context that I'm not aware of.  What I can say is that I've been running my fork of Y.js (the one the PR is based on) for months and it has worked perfectly in every browser I've tried with documents up to about 60,000 words.

fsctl:
This PR fixes a bug in `cleanupTransactions()` that causes the call stack to overflow when loading a large Y.Doc.  The fix is to change that function from a recursive function to an iterative function so that it can never overflow the call stack no matter how many cleanups are needed.

There is a full description of the bug in #522.

Closes #522

### DIFF
--- a/src/utils/Transaction.js
+++ b/src/utils/Transaction.js
@@ -242,7 +242,7 @@ export const tryGc = (ds, store, gcFilter) => {
  * @param {number} i
  */
 const cleanupTransactions = (transactionCleanups, i) => {
-  if (i < transactionCleanups.length) {
+  while (i < transactionCleanups.length) {
     const transaction = transactionCleanups[i]
     const doc = transaction.doc
     const store = doc.store
@@ -363,11 +363,9 @@ const cleanupTransactions = (transactionCleanups, i) => {
         subdocsRemoved.forEach(subdoc => subdoc.destroy())
       }
 
-      if (transactionCleanups.length <= i + 1) {
+      if (transactionCleanups.length <= ++i) {
         doc._transactionCleanups = []
         doc.emit('afterAllTransactions', [doc, transactionCleanups])
-      } else {
-        cleanupTransactions(transactionCleanups, i + 1)
       }
     }
   }


dmonad:
Please see my comment in https://github.com/yjs/yjs/issues/522#issuecomment-1512810792

PR #528, opened 2023-04-30
Add missing getting-started steps

jamesgpearce:
The 'getting started' doesn't work without importing and creating the first doc.


dmonad (approved):
Makes sense, thank you!

Issue #529, opened 2023-05-03
Text deletion event includes erroneous retain attribute


ezeidman:
**Describe the bug**
Deleting text can lead to a YEvent which includes an extra attribute update

**To Reproduce**
1. Create an XmlText which contains in order (1) some text with formatting and (2) an embedded XmlText
2. observeDeep events on the parent XmlText
3. Delete the text prefix from the XmlText

Or in code:
```
  const doc =  new Y.Doc()
  const root = doc.get("text", Y.XmlText) as Y.XmlText;
  const p1 = new Y.XmlText();
  p1.setAttribute("type", "paragraph")
  root.insertEmbed(0, p1);
  root.insert(0, "Text", {bold: true});

  root.observeDeep((events) => {
    events.forEach(event => {
      console.log("Event", JSON.stringify(event.changes, null, 4));
    })
  })
  root.delete(0, 4);
  ```

**Expected behavior**
Expect to observe an event with a delta that has one delete with a length equal to however long the text was.

**Actual behavior**
For some reason the event includes a retain that tries to null out the formatting attribute on the embedded XmlText

<img width="224" alt="Screenshot 2023-05-02 at 6 53 04 PM" src="https://user-images.githubusercontent.com/1455518/235819591-d9392744-4ceb-4b71-b9e8-85e7f4391b5c.png">

**Environment Information**
 - Yjs version = 13.5.45

**Additional context**
Ran into this problem while using Slate Yjs, which turns the retain into set_node Slate operation over here: https://github.com/BitPhinix/slate-yjs/blob/966b9325ff312058023d017f857fbb6ac9a7a8a4/packages/core/src/applyToSlate/textEvent.ts#L88. This then tries to update the representation of the paragraph in Slate based on formatting attributes that the text had, which doesn't really make sense and leads to an invalid state.

There probably are a couple workarounds for this issue but it seems odd that deleting text would result in a retain signaling a change to content that didn't actually change. Is there any reason to structure the YEvents this way?

From debugging the issue it seems when we're creating the YTextEvent, when we process the first deleted ContentFormat, we set the attribute to null in `attributes` but we don't remove it when processing the second deleted ContentFormat. This is around https://github.com/yjs/yjs/blob/e0a2f11db325ae570ca273c4e75a1b9e8655491a/src/types/YText.js#L718-L727

ezeidman:
[...]

PR #530, opened 2023-05-03
Fix retains for embedded types in YTextEvent

ezeidman:
Don't include attributes in retains for embedded types since they're not subject to formatting

This fixes https://github.com/yjs/yjs/issues/529

### DIFF
--- a/tests/y-text.tests.js
+++ b/tests/y-text.tests.js
@@ -1676,6 +1676,29 @@ export const testDeltaAfterConcurrentFormatting = tc => {
 /**
  * @param {t.TestCase} tc
  */
+export const testDeltaAfterDelete = tc => {
+  const { text0, text1, testConnector } = init(tc, { users: 2 })
+  text0.insert(0, 'abc', {bold: true});
+  text0.insertEmbed(3, new Y.Map([['key', 'val']]));
+  testConnector.flushAllMessages()
+  /**
+   * @type {any}
+   */
+  const deltas = []
+  text1.observe(event => {
+    if (event.delta.length > 0) {
+      deltas.push(event.delta)
+    }
+  })
+  text0.delete(0, 3);
+  testConnector.flushAllMessages()
+  console.error("The actual deltas", deltas);
+  t.compare(deltas, [[{ delete: 3}]])
+}
+
+/**
+ * @param {t.TestCase} tc
+ */
 export const testBasicInsertAndDelete = tc => {
   const { users, text0 } = init(tc, { users: 2 })
   let delta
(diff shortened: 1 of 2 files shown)


dmonad:
Thank you! 
Can you please confirm that this is the same behavior as in https://github.com/quilljs/delta/ ?

ezeidman:
> Thank you! Can you please confirm that this is the same behavior as in https://github.com/quilljs/delta/ ?

Could you clarify what you mean? The deltas from both before and after the PR are valid Quill deltas. The issue however is that retains associated with embedded types (e.g. an XmlText inside an XmlText) can contain unrelated attribute updates that are actually associated with surrounding text. This could be an erroneous null attribute like the example in the linked ticket or an erroneous non-null attribute value. 

An example:

```
  const doc =  new Y.Doc()
  const root = doc.get("text", Y.XmlText) as Y.XmlText;
  const p1 = new Y.XmlText();
  p1.setAttribute("type", "paragraph")
  p1.insert(0, "p1")
  
  root.insert(0, "abc", {bold: true});
  root.insert(3, "defghi", {bold: false});
  root.insertEmbed(6, p1);
  
  root.observeDeep((events) => {
    events.forEach(event => {
      console.log("Event", JSON.stringify(event.changes, null, 4));
    })
  })

  root.delete(3, 3);
```

produces
Screenshot 2023-05-05 at 2 28 00 PM

where the retain is applying bold=false to the XmlText even though it's not bold and neither are its contents.

There is one issue with this PR as is: it can create more retains than are strictly necessary. For example [retain 3, retain 4, retain 1], instead of just retain 8. That's still a valid Quill delta but not the most concise way to represent things.

tests/y-text.tests.js
```diff
@@ -1673,6 +1673,82 @@ export const testDeltaAfterConcurrentFormatting = tc => {
...
+   */
+  const deltas = []
+  text1.observe(event => {
+    if (event.delta.length > 0) {
+      deltas.push(event.delta)
+    }
+  })
+  text0.delete(3, 3);
+  testConnector.flushAllMessages()
+  t.compare(deltas, [[{retain: 3}, { delete: 3}]])
```
ezeidman, line 1727: Before the fix this is `[[{"retain":3},{"delete":3},{"retain":4,"attributes":{"bold":false}}]]`

tests/y-text.tests.js
```diff
@@ -1673,6 +1673,82 @@ export const testDeltaAfterConcurrentFormatting = tc => {
...
+   */
+  const deltas = []
+  text1.observe(event => {
+    if (event.delta.length > 0) {
+      deltas.push(event.delta)
+    }
+  })
+  text0.delete(0, 3);
+  testConnector.flushAllMessages()
+  t.compare(deltas, [[{ delete: 3}]])
```
ezeidman, line 1700: Before the fix this is `[[{"delete":3},{"retain":1,"attributes":{"bold":null}}]]`

PR #539, opened 2023-06-15
Feature: WeakLink

Horusiath:
This PR introduces new `WeakLink` primitive, which can be used to create references to data residing in other parts of the document tree (therefore making it into a graph). Its API is similar to JavaScript [WeakRef](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef). 

1. You can create links to map and array elements via `ymap.link('key')`/`yarray.link(index)`. I didn't make it for XML types, as it doesn't have any sense: XML is acyclic tree-like structure by definition. You cannot make links referencing to fragments of Text type (at the moment at least), but you should be able to embed them inside Text.
2. Underlying link values can be retrieved via `ylink.deref()` method just like in JavaScript WeakRef. If linked value has been deleted, `undefined` will be returned.
3. Links can point to scalar values or other shared y-types.
4. Deleting link WON'T delete linked value.
5. Links can be observed on. Map link changes are automatically propagated as key is being updated or deleted. Array links changes react on deletes, but inserting new value at the same index DOESN'T count as updating the link's value.
6. Links can be used in combination with deep observe calls. This way changes in y-types are propagated not only up in their parent hierarchy, but also spread out across their links (and further up to link's parents). If such hierarchy creates cycles, they will be caught and prevent from making stack overflow.


src/types/YWeakLink.js
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  /**
+   * @param {UpdateEncoderV1 | UpdateEncoderV2} encoder
+   */
+  _write (encoder) {
+    encoder.writeTypeRef(YWeakLinkRefID)
+    const flags = 0 // flags that could be used in the future
+    encoding.writeUint8(encoder.restEncoder, flags)
+    encoder.writeLeftID(this._id)
```
Horusiath, line 137: Encoding scheme is basically: 
1. `0` (1 byte) - flag that can be used in the future (eg. for links spanning over ranges rather than single elements etc.)
2. ID of linked element.

I'm not sure about `writeLeftID` here, as it will almost surely mess up with lib0 v2 delta-of-deltas encoding. Maybe just write ID into restEncoder?

tests/y-weak-links.tests.js
```diff
@@ -0,0 +1,475 @@
...
+  nested.set('key', 'value2')
+  t.compare(events.length, 1)
+  t.compare(events[0].target, nested)
+  t.compare(events[0].keys, new Map([['key', {action:'update', oldValue: undefined}]]))
+
+  // delete entry in linked map
+  nested.delete('key')
+  t.compare(events.length, 1)
+  t.compare(events[0].target, nested)
+  t.compare(events[0].keys, new Map([['key', {action:'delete', oldValue: undefined}]]))
```
Horusiath, line 376: This assertion fails - moreover it'll also fail in following example (deep observe over nested maps):

```js
const doc = new Y.Doc()
const outer = doc.getMap('outer')
let events = []
outer.observeDeep((e) => events = e)

const inner = new Y.Map()
outer.set('inner', inner) // clock: 1

inner.set('key', 'value1') // clock: 2
inner.set('key', 'value2') // clock: 3
inner.delete('key')
  
t.compare(events[0].keys, new Map([['key', {action:'delete', oldValue: undefined}]]))
```

I think it's a bug. The reason for that is that `event.keys` property is computed by checking if corresponding block has been added/deleted within the scope of the transaction. 

However what happens here is that we subsequently set "key" to "value1" (item clock: 1, length: 1), then "value2" (clock: 2, length: 1). When we delete "key", these two blocks get squashed and the transaction's delete set can no longer recognise squashed block as deleted (as only fragment of that block was deleted in current transaction), which in result causes `event.keys` to skip over the deleted key.

src/types/YWeakLink.js
dmonad: Yeah, I think that would be better

tests/y-weak-links.tests.js
dmonad: You are gonna love this.

`event.change` is a computed property that calculates the values dynamically. So if you call it outside of the event handler (i.e. after the transaction cleanup) you're gonna have a bad time.

If you call `events[0].keys` in the observer, the value is properly cached.

I will add an error message when you do that.

src/structs/Item.js
```diff
@@ -295,6 +296,13 @@ export class Item extends AbstractStruct {
      * @type {ID | null}
      */
     this.redone = null
+    /**
+     * If this item was referenced by other weak links, here we keep the references
+     * to these weak refs.
+     * 
+     * @type {Set<YWeakLink<any>> | null}
+     */
+    this.linkedBy = null
```
dmonad, line 305: This will increase the size of all Item objects. This should be stored on AbstractType instead.

src/structs/Item.js
```diff
@@ -376,6 +384,13 @@ export class Item extends AbstractStruct {
     if (this.parent && this.parent.constructor === ID && this.id.client !== this.parent.client && this.parent.clock >= getState(store, this.parent.client)) {
       return this.parent.client
     }
+    if (this.content.constructor === ContentType && /** @type {ContentType} */ (this.content).type.constructor === YWeakLink) {
```
dmonad, line 387: Is this necessary?

If this is a WeakLink, then it could also exist when the type does not exist yet.

Two reasons to remove this condition:
1. This can (potentially) lead to circular dependencies that can't be resolved (`ymap->yweaklink(yarray)` + `yarray->yweaklink(ymap)`).
2. It enables us to possibly extend WeakLinks to types of subdocuments that are eventually loaded (I'm definitely not suggesting to implement this right now)

src/structs/Item.js
```diff
@@ -578,6 +600,8 @@ export class Item extends AbstractStruct {
       this.deleted === right.deleted &&
       this.redone === null &&
       right.redone === null &&
+      this.linkedBy === null &&
```
dmonad, line 603: Types (ContentType) can't be merged. If the type is deleted, we can simply forget the reference.

src/types/YArray.js
```diff
@@ -201,6 +203,22 @@ export class YArray extends AbstractType {
...
   }
 
+  /**
+   * Returns the weak link to i-th element from a YArray.
+   *
+   * @param {number} index The index of the element to return from the YArray
+   * @return {YWeakLink<T>}
+   */
+  link(index) {
+    if (this.doc !== null) {
```
dmonad, line 213: I propose to limit WeakLinks only to refer to types. This would simplify the implementation.

Mainly, I want to avoid having a list of WeakLinks references on each type. Alternatively, WeakLink could only store the ID and look up the item every time the value is requested (if the item is deleted, then we return undefined).

src/types/YWeakLink.js
```diff
@@ -0,0 +1,206 @@
...
+ * @extends YEvent<any>
+ * Event that describes the changes on a YMap.
+ */
+export class YWeakLinkEvent extends YEvent {
+  /**
+   * @param {YWeakLink<T>} ylink The YWeakLink to which this event was propagated to.
+   * @param {Transaction} transaction
+   */
+  constructor (ylink, transaction) {
+    super(ylink, transaction)
```
dmonad, line 23: Idea: If this refers to a type, it could contain the event of the referred type (if it was changed) and also all deep events generated on the referred type (`tr.changedParents.get(referredType)`).

PR #540, opened 2023-06-15
Ytext cleanup 538 refactor

dmonad:
Based on #538

### DIFF
--- a/src/utils/Transaction.js
+++ b/src/utils/Transaction.js
@@ -115,9 +115,9 @@ export class Transaction {
      */
     this.subdocsLoaded = new Set()
     /**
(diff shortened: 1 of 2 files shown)


src/utils/Transaction.js
```diff
@@ -295,6 +300,9 @@ const cleanupTransactions = (transactionCleanups, i) => {
         fs.push(() => doc.emit('afterTransaction', [transaction, doc]))
       })
       callAll(fs, [])
+      if (transaction._needFormattingCleanup) {
```
dmonad, line 303: if (needFormattingCleanup && beforeState !== {}) @dmonad

PR #542, opened 2023-06-16
mark structs as merged and filter out at the end of bulk operations

### LINKED ISSUE #541: Slow apply when garbage collecting large update
**Checklist**

**Describe the bug**
Given a large, ungarbagecollected doc which was persisted as an encoded update, if you apply the update to a document with garbage collection enabled, the apply can take a very long time.
Based on my profiling, the time is all in [tryToMergeWithLeft](https://github.com/yjs/yjs/blob/main/src/utils/Transaction.js#L170), possibly due to calling `structs.splice(pos, 1)` many times, requiring the shifting of a very large array each time.

**To Reproduce**
A anonymized document which reproduces this problem can be found here:
https://github.com/NilSet/yjs-stack-overflow/tree/slow-gc
On my machine it takes 38 seconds to apply the document with garbage collection enabled.
Note that if you change the test script to disable gc on the doc, it only takes 1.6 seconds to apply the document.

**Expected behavior**

**Screenshots**

**Environment Information**
 - Node 18
 - Yjs 13.6.4

NilSet:
instead of splicing out of the array immediately
Fixes #541

### DIFF
--- a/src/utils/StructStore.js
+++ b/src/utils/StructStore.js
@@ -112,6 +112,9 @@ export const addStruct = (store, struct) => {
 export const findIndexSS = (structs, clock) => {
   let left = 0
   let right = structs.length - 1
+  while (/** @type any */(structs[right])._wasMerged) {
+    right--
+  }
   let mid = structs[right]
   let midclock = mid.id.clock
   if (midclock === clock) {
@@ -126,6 +129,9 @@ export const findIndexSS = (structs, clock) => {
     midclock = mid.id.clock
     if (midclock <= clock) {
       if (clock < midclock + mid.length) {
+        while (/** @type any */(structs[midindex])._wasMerged) {
+          midindex--
+        }
         return midindex
       }
       left = midindex + 1
--- a/src/utils/StructStore.js
+++ b/src/utils/StructStore.js
@@ -112,7 +112,7 @@ export const addStruct = (store, struct) => {
 export const findIndexSS = (structs, clock) => {
   let left = 0
   let right = structs.length - 1
-  while (/** @type any */(structs[right])._wasMerged) {
+  while (right > 0 && structs[right] === structs[right - 1]) {
     right--
   }
   let mid = structs[right]
@@ -129,7 +129,7 @@ export const findIndexSS = (structs, clock) => {
     midclock = mid.id.clock
     if (midclock <= clock) {
       if (clock < midclock + mid.length) {
-        while (/** @type any */(structs[midindex])._wasMerged) {
+        while (midindex > 0 && structs[midindex] === structs[midindex - 1]) {
           midindex--
         }
         return midindex
--- a/src/utils/Transaction.js
+++ b/src/utils/Transaction.js
@@ -184,24 +184,21 @@ const tryToMergeWithLeft = (structs, pos) => {
  * @param {Array<AbstractStruct>} structs
  * @param {number} pos
  */
-const tryToMergeWithLeftNoSplice = (structs, pos) => {
+export const tryToMergeWithLeftNoSplice = (structs, pos) => {
+  const left = structs[pos - 1]
   const right = structs[pos]
-  if (/** @type any */(right)._wasMerged) {
+  if (left === right) {
     return 0
   }
-  let offset = 1
-  let left
-  do {
-    left = structs[pos - offset++]
-  } while (/** @type any */(left)._wasMerged)
   if (left.deleted === right.deleted && left.constructor === right.constructor) {
     if (left.mergeWith(right)) {
-      /** @type any */(right)._wasMerged = true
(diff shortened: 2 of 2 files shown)


src/utils/Transaction.js
```diff
@@ -180,6 +180,48 @@ const tryToMergeWithLeft = (structs, pos) => {
...
+    }
+  }
+  return 0
+}
+
+/**
+ * @param {Array<AbstractStruct>} structs
+ * @param {number} mergedCount
+ */
+const filterMergedStructs = (structs, mergedCount) => {
```
NilSet, line 213: probably better off constructing a new array using array.filter but im not sure if there will be references to the old array dangling so gotta do this ugly mutation

src/utils/StructStore.js
```diff
@@ -126,6 +126,9 @@ export const findIndexSS = (structs, clock) => {
     midclock = mid.id.clock
     if (midclock <= clock) {
       if (clock < midclock + mid.length) {
+        while (/** @type any */(structs[midindex])._wasMerged) {
```
NilSet, line 129: im pretty sure this function is not quite correct yet

NilSet: fixed the early return case returning one of the ghost items

en-tcelvi:
I'll admit, I don't full understand what is going on in this PR, but would it be possible to do something like this:

```
**
 * @param {Array<AbstractStruct>} structs
 * @param {number} pos
 */
const tryToMergeWithLeft = (structs, pos) => {
  const left = structs[pos - 1]
  const right = structs[pos]
  if (left.deleted === right.deleted && left.constructor === right.constructor) {
    if (left.mergeWith(right)) {
      if (right instanceof Item && right.parentSub !== null && /** @type {AbstractType<any>} */ (right.parent)._map.get(right.parentSub) === right) {
        /** @type {AbstractType<any>} */ (right.parent)._map.set(right.parentSub, /** @type {Item} */ (left))
      }
      return true;
    }
  }
  return false;
}

const tryToMergeWithLeftProvider = () => {
  /**
   * @type {Array<number[]>}
   */
  const spliceRanges = []
  spliceRanges.push([0])
  return {
    /**
     * @param {Array<AbstractStruct>} structs
     * @param {number} pos
     */
    exec: (structs, pos) => {
      const needsSplice = tryToMergeWithLeft(structs, pos)
      const spliceRange = spliceRanges[spliceRanges.length - 1]
      if(needsSplice) {
        spliceRange[1] = pos
        spliceRange[0]++
      } else if (spliceRange.length === 2) {
        // @ts-ignore
        spliceRange.push([0]);
      }
    },
    /**
     * @param {Array<AbstractStruct>} structs
     */
    flush: (structs) => {
      for (const spliceRange of spliceRanges) {
        if (spliceRange.length === 2) {
          structs.splice(spliceRange[1], spliceRange[0])
        }
      }
    }
  };
```

Usage:

```
const tryMergeDeleteSet = (ds, store) => {
  // try to merge deleted / gc'd items
  // merge from right to left for better efficiecy and so we don't miss any merge targets
  ds.clients.forEach((deleteItems, client) => {
    const structs = /** @type {Array<GC|Item>} */ (store.clients.get(client))
    for (let di = deleteItems.length - 1; di >= 0; di--) {
      const deleteItem = deleteItems[di]
      // start with merging the item next to the last deleted item
      const mostRightIndexToCheck = math.min(structs.length - 1, 1 + findIndexSS(structs, deleteItem.clock + deleteItem.len - 1))
      const mergeLeftProvider = tryToMergeWithLeftProvider()
      for (
        let si = mostRightIndexToCheck, struct = structs[si];
        si > 0 && struct.id.clock >= deleteItem.clock;
        struct = structs[--si]
      ) {
        mergeLeftProvider.exec(structs, si)
      }
    }
  })
}
```

The idea being we track the data in another data structure than transact as opposed to adding values to the struct as we process them?  (Fully aware that might sample code does not pass tests - just trying contribute to the iteration of the solution somehow)

NilSet:
Conceptually, while merging structs, a struct can become "wider" in the struct store, taking up multiple array indexes. The binary search accounts for this, and tries to return the left-most index of the wide struct.

We scan the array at the end to shrink the wide structs into a single position in the array.

en-tcelvi:
@dmonad Can we bump the priority on this one please?

dmonad:
Thanks @NilSet! I spent the week and thought more about this problem.

This PR adds special behavior to how we interact with the data structure that stores the structs. In Yjs, we use a simple Array. In Yrs, we created a special data structure. Eventually, I want to do the same in Yjs. Hence, I feel we shouldn't add additional complexity specific to StructStore using Arrays & binary search. It should be possible to keep the data structure in a valid state without duplicates. A better data structure will eventually avoid splicing altogether.

I distilled the idea to a simpler fix #553 (admittedly covering fewer cases). It also avoids iterating again through the structs to filter out the duplicates. This should fix #541 as well.

PR #543, opened 2023-06-17
fix: Improve the getAttributes method implementation

WillBean:
Our ProseMirror project uses the `renderSnapshot` API to implement the history comparison feature. However, I found that the `getAttrbutes` method is called in the [sync-plugin file](https://github.com/yjs/y-prosemirror/blob/master/src/plugins/sync-plugin.js#L662), which obtains attributes based on the snapshot.It seems that this method has not been fully implemented, so I have completed its implementation.

### DIFF
--- a/src/types/AbstractType.js
+++ b/src/types/AbstractType.js
@@ -877,19 +877,28 @@ export const typeMapGet = (parent, key) => {
 
 /**
  * @param {AbstractType<any>} parent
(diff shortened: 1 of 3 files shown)


dmonad:
Thanks @WillBean,

The snapshot API is quite delicate. I implemented the suggested feature in the commit mentioned above. I'll make a new release in a bit.

PR #545, opened 2023-06-19
invalidate cached path when changing currentTarget of event

### LINKED ISSUE #544: yjs 13.6.4 breaks paths in observed events
This test case fails with 13.6.4, but succeeds with 13.6.3 and 13.6.1:
```typescript
import Y from 'yjs';
import assert from 'node:assert/strict';

describe('Task bug on YJS', () => {
  it('YMapEvent + YTextEvent paths', () => {
    const doc = new Y.Doc();
    const map1 = doc.getMap('tasks');
    const task1Map = new Y.Map();
    const textNode = new Y.Text('my task label');
    task1Map.set('label', textNode);
    map1.set('task1', task1Map);

    // Listen to events
    const loggedPaths = [];
    map1.observeDeep((events, transaction) => {
      console.log('Begin transaction');
      for (const event of events) {
        const eventType =
          event instanceof Y.YMapEvent ? 'YMapEvent' : event instanceof Y.YTextEvent ? 'YTextEvent' : 'Unknown';
        console.log(`${eventType}, path: ${event.path}`);

        loggedPaths.push(event.path as any);
      }
      console.log('End transaction\n');
    });

    doc.transact(() => {
      textNode.insert(0, 'my task label - ');
      task1Map.set('meta1', new Y.Text('meta1111'));
    });

    // On 13.6.4, loggedPaths == [[], ['label']]
    assert.deepEqual(loggedPaths, [['task1'], ['task1', 'label']]);
  });
});
```

My hypothesis is the following: in 13.6.4, `YEvent.path` getter was changed to cache the value. However, in `Transaction.js` a line of codes [mutates](https://github.com/yjs/yjs/blob/ce098d0ac2b617d0f4c5214c95607dee8cfaba87/src/utils/Transaction.js#L289) `YEvent.currentTarget`, one of the parameters used to compute the path. So, it is not possible to cache that value safely.

The bug happens when sibling `YMap` and `YText` are changed in the same transaction.

**To Reproduce**
Run the test first on 13.6.3 and verify that it passes.
Run it again on 13.6.4, and it should fail on the final assert.

**Expected behavior**
The test should pass on 13.6.4.

**Environment Information**
 - NodeJS 18.16.0
 - YJS 13.6.4

NilSet:
fixes #544

### DIFF
--- a/src/utils/Transaction.js
+++ b/src/utils/Transaction.js
@@ -287,6 +287,7 @@ const cleanupTransactions = (transactionCleanups, i) => {
               events
                 .forEach(event => {
                   event.currentTarget = type
+                  event._path = null
                 })
               // sort events by path length so that top-level events are fired first.
               events
--- a/tests/y-map.tests.js
+++ b/tests/y-map.tests.js
@@ -337,6 +337,34 @@ export const testObserversUsingObservedeep = tc => {
   compare(users)
 }
 
+/**
+ * @param {t.TestCase} tc
+ */
+export const testPathsOfSiblingEvents = tc => {
+  const { users, map0 } = init(tc, { users: 2 })
+  /**
+   * @type {Array<Array<string|number>>}
+   */
+  const pathes = []
+  let calls = 0
+  const doc = users[0]
+  map0.set('map', new Y.Map())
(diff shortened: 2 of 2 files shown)


dmonad:
Thank you so much!

PR #548, opened 2023-06-21
fix equalDeleteSets

YousefED:
`equalDeleteSets` didn't work properly, because the `return false` in the foreach only quits a "single foreach", not the entire function


dmonad:
oh my .. thanks @YousefED !

equalDeleteSets didn't work properly, because the return false in the foreach only quits a "single foreach", not the entire function

PR #556, opened 2023-07-23
Update README.md

AdventureBeard:
Adding LegendKeeper to the "Yjs in the wild" list.


dmonad:
Of course, thanks for adding it!

Issue #558, opened 2023-08-11
mergeUpdates incorrect result


himself65:
> Please save me some time and use the following template. In 90% of all issues I can't reproduce the problem because I don't know what exactly you are doing, in which environment, or which y-* version is responsible. Just use the following template even if you think the problem is obvious.

**Checklist**

* [x] Are you reporting a bug? Use github issues for bug reports and feature requests. For general questions, please use https://discuss.yjs.dev/
* [x] Try to report your issue in the correct repository. Yjs consists of many modules. When in doubt, report it to https://github.com/yjs/yjs/issues/

**Describe the bug**

**To Reproduce**
I don't know how to create mini reproduce

**Expected behavior**
Works correctly

**Screenshots**
![64f558c2-7fef-4086-a857-82043d4f000d](https://github.com/yjs/yjs/assets/14026360/ef95a9a0-df7c-4e43-a68d-806182ca5f55)

**Environment Information**
 - Chrome
 - 13.6.7

**Additional context**
Related PR: https://github.com/toeverything/AFFiNE/pull/3687

himself65:
this.left, this.origin, this.right will be `undefined` on the clock `0` for some reason after `mergeUpdates`

himself65:
The weird thing is when I try to reproduce the same data in node.js. It works perfectly.

Let me do the same thing in the browser later

himself65:
![ef9b59e9-95ad-45ac-89a7-d8163e31766c](https://github.com/yjs/yjs/assets/14026360/6359acaf-679a-4bf7-a3c3-70a6bcc60185)

I think the correct type here is `GC | Item | null`

![image](https://github.com/yjs/yjs/assets/14026360/7366ca07-57a6-4cfc-a2da-567193dfb28f)

dmonad:
I commented on your PR as well. Could you send me a reproducible sample?

It would also help if you'd send me a stacktrace of the error.

I don't understand how we can end up in this situation. `left` should never be garbage-collected. Since I only saw minified code producing this issue, it might actually be the minifier that produces bad code by, for example, misinterpreting some condition.

dmonad:
The properties of `parent` would also be interesting to me.

himself65:
I debugged it last week. Seems like the error happened in `client: 748511660` and `clock: 0`

himself65:
> It's about 24MB.
> 
> [1.txt](https://github.com/yjs/yjs/files/12456154/1.txt)

[...]

PR #559, opened 2023-08-11
fix: left might be GC

himself65:
This case happened in `mergeUpdates` API. I'm not sure how to reproduce this, but I'm pretty sure `this.left` could be `GC`

Fixes: https://github.com/yjs/yjs/issues/558


himself65:
I think the correct solution is fix mergeUpdates

dmonad:
I received a dataset that helped me to find the issue and fix it. I will publish a new release shortly.

PR #564, opened 2023-08-23
add oorja.io to who is using yjs

akshaykmr:
I've been using yjs since 2017. Excellent library.
oorja.io is my side-project that refuses to die. It uses the yjs doc for participant collaboration and persistence in apps like codeshare, excalidraw within.

Cheers


dmonad:
Thanks, I'm glad to hear that! 

Love the UI :)

PR #566, opened 2023-08-28
Adds IllumiDesk as a reference to the using list

jgwerner:
Adds [IllumiDesk](https://illumidesk.com) as a reference user for Yjs.


dmonad:
Great, thanks!

PR #575, opened 2023-09-15
fix clientID description (53bit to 32bit)

nikgraf:
Initially I thought we can switch to the 53 space and opened a PR https://github.com/yjs/yjs/pull/572, but then realized that this breaks Awareness since there the assumption is that the `clientID` is 32bits due the use of `readVarUint`


dmonad:
Unfortunately, there is a slight difference in "how it should be" and "how it currently is". But 53bit describes "how it should be" and where we will go. 53bit should already be fully supported.

`readVarUint` should be able to handle 53bit numbers (unless you are on an older release of lib0). I will bump the `lib0` explicitly in y-protocols so this doesn't happen. Please reopen this ticket if the issue persists!

PR #580, opened 2023-10-06
Add Evernote as a user of Yjs

gveres:
According to https://evernote.com/blog/future-proofing-evernotes-foundations/ they have moved to Yjs lately


dmonad:
Right, thanks!

PR #583, opened 2023-10-07
add documentation on V2 events

nikgraf:
Multiple people I talked to didn't figure this out and therefor sticked to V1. I hope this should make it easier for others to upgrade.


holtwick:
Thanks @nikgraf

raineorshine:
Fwiw, `applyUpdate` is exported [by default](https://github.com/yjs/yjs/blob/da8ca5168e089d64150ce1e8d3ee863b4e5dd7d6/src/utils/encoding.js#L493), so I'm not sure it's good to advertise v1 vs v2 anymore.

dmonad:
Thanks Nik!

@raineorshine  Switching to V2 would be a breaking change. Now that y-crdt supports v2 as well, we could make V2 the default. But I want to do that in the next major release.

PR #593, opened 2023-10-31
fix: remove unused if-statement check

himself65:
I'm going through the source code of yjs and find some unused logic. This might improve a little performance


dmonad:
Right, thanks!

PR #594, opened 2023-11-01
fix: use switch instead of ifs

himself65:


dmonad:
This is a fun approach to use switch instead of ifs! I haven't seen this yet..

Style: :100: 

I'm gonna reject this PR only because it doesn't necessarily improve clarity, but it does slightly increase the character count.

PR #595, opened 2023-11-06
feat: expose some types

himself65:
I'm trying to implement some end-to-end encryption updates, like `encryptUpdate(update, private) -> update,` that allows the server to merge updates and decode updates...(just like `obfuscateUpdate`), but the difference is contents are encrypted by `privateKey.` So that it can be safely saved to the server, but the update is still valid so that the server can validate if the document schema is correct.

Then, I found that some internal classes are not exposed, so it will make things hard to extend the existing logic. I would appreciate it if some internal types could be exposed to the downstream so that we can make the ecosystem better.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -18,8 +18,10 @@ export {
   Item,
   AbstractStruct,
   GC,
+  Skip,
   ContentBinary,
   ContentDeleted,
+  ContentDoc,
   ContentEmbed,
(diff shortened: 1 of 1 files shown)


himself65:
@dmonad any idea of this PR?

dmonad:
Sure, gonna make a release soon!

PR #596, opened 2023-11-07
Adding www.btw.so to the platforms who use Yjs

siddug:
Repo here: https://github.com/btw-so/btw


dmonad:
Awesome, thanks for sharing!

PR #600, opened 2023-11-16
Add support for async functions inside `doc.transact`

remorses:
Transact will wait for the async function passed to `transact` to complete before cleaning up the transaction

My use case is streaming text words 1 by 1 with some exceptions were i need to group many updates in 1, without support for async functions inside transactions this is impossible

### DIFF
--- a/src/utils/Transaction.js
+++ b/src/utils/Transaction.js
@@ -436,7 +436,13 @@ export const transact = (doc, f, origin = null, local = true) => {
         // observes throw errors.
         // This file is full of hacky try {} finally {} blocks to ensure that an
         // event can throw errors and also that the cleanup is called.
-        cleanupTransactions(transactionCleanups, 0)
+        if (typeof result?.then === 'function' && result?.constructor?.name === 'Promise') {
+          result.then(() => {
+            cleanupTransactions(transactionCleanups, 0)
+          })
+        } else {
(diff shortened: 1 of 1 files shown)


dmonad:
Sorry @remorses, this goes against the design concept of Yjs. This will break existing editor bindings, once you start using async transactions.

1. document transactions must complete immediately after calling `type.insert(..)`, so that editor bindings receive changes before they themselves mutate the document.
4. This would break all existing editor bindings, once you start using async transactions.
3. This approach will basically block all "normal" transactions while the async transaction is active.
4. I don't see a real use-case for this yet / it doesn't solve a problem that I understand. Streaming characters can also be done in non-async transactions without breaking editor binings. 
5. It is unclear how this approach would interact with the Y.UndoManager.
6. I associate Items to transactions by storing ranges of lamport timestamps (clock `n` to `m`) on the transaction. However, that won't work if there can be concurrent transactions generating lamport timestamps while an async transaction is still "running".
7. synchronousity is awesome, we should embrace it. 

In short, a feature like this (even if it would solve a problem) would require a lot more design effort to don't break existing stuff.

Cheers,
Kevin

remorses:
Thank you for the detailed explanation, I solved this by buffering async updates and grouping them in a single transaction

My specific use case was streaming table updates, you need to create the rows cells synchoronously or prosemirror would automatically add cells for you

PR #601, opened 2023-11-25
(bug): Check if structs are empty before findIndexSS and writing encoding

websiddu:
When applying a snapshot to a brand new yDoc, on the server the state is not available, so the original doc is fresh. In such case just ignore writing the structs.


dmonad:
Hi @websiddu,

First, can you please create some kind of example that we can use to verify that this works?  Ideally, you create a new test-case.

The PR seems incomplete to me, because if you don't write a set of structs, then you also need to adjust the `size` which is written before.

PR #605, opened 2023-12-23
Update README.md with mobx-keystone binding

xaviergonz:
As the name says, a new binding for the list :)


dmonad:
Sweet, thanks!

PR #609, opened 2024-02-05
Add top-level getXmlElement

raineorshine:
All shared types have top-level getters except `YXmlElement`. 

This PR adds `getXmlElement` to the `Doc` class.

```js
const doc = new Y.Doc()
const yxmlel = ydoc.getXmlElement('prop-name')
```

Reference:

- https://discuss.yjs.dev/t/why-does-the-ydoc-getxmlelement-method-not-exist/2249


dmonad:
Thank you @raineorshine !

I also added better typings for `ydoc.get`. It makes the bracket part in the following snipped obsolete: `ydoc.get('name', Y.XmlElement) [as Y.XmlElement]`.

I will add this to the next release.

PR #614, opened 2024-02-25
Improve types for map public interface

### LINKED ISSUE #499: Incorrect types for Map.get & Map.set
**Checklist**

**Describe the bug**
When using yjs with Typescript, there doesn't seem to be any typesafe way to use `Map.get` or `Map.set`. `Map` objects require a `MapShape` type, but the return type of `Map.get("key")` and the new value type for `Map.set("key", value)` are both `MapShape | undefined`, which doesn't match the behavior of those methods.

**To Reproduce**
In a typescript project:

```ts
type Item = {
  id: number;
  name: string;
}

const map = Y.Map<Item>();
map.get("name") // Should be 'string | undefined'
map.get("id") // Should be 'number | undefined'
map.set("name", "my cool name") // Throws a type error becaus the second argument is not an Item
```

**Expected behavior**
I would just fix this myself, but yjs uses a type generation method I'm unfamiliar with (JSDoc?).

Here are the types I would roughly expect in TS syntax:

```ts
class Map<MapShape> {
  get(key: keyof MapShape): MapShape[typeof key] | undefined;
  set(key: keyof MapShape, value: MapShape[typeof key]): typeof value;
  delete(key: keyof MapShape): void
}
```

I don't know whether this specific form of type inference is supported by your type generator, but at the very least the types shouldn't make it impossible to write code without type errors.

**Environment Information**
 - latest yjs
 - TS version 4.4.4

Smona:
Resolves #499. 

This type-only PR modifies the semantics of `Y.Map`'s generic type parameter, to make it describe the exact shape of the map rather than the type of its values. This parameter is then used to improve the argument and return types of public-facing `Map` methods, where it makes sense to do so.

### Model schemas

This change enables the use of the type parameter to validate the types of maps which represent "models", which may have values of different types and only certain valid keys. Currently, users trying to use YJS as a frontend database for TypeScript will frequently run into this issue:

```ts
type User = {
  name: string;
  age?: number;
};

const map = doc.getMap<User>();

// ERROR: Argument of type 'string' is not assignable to parameter of type 'User'
user.set("name", "Tina Belcher")

// userAge has the incorrect type `User | undefined`
const userAge = map.get("age")
```

As someone who writes a lot of TS, these issues significantly hamper adoption in complex software.

### Helping maintain data consistency

Even though YJS has a distributed data model, many designs will limit contributions to a particular YJS document to different users of the same software. By presenting type errors when attempting to set a field to a type that does not match the developer's specified schema, we can help developers to protect document consistency and prevent potentially tricky to debug document poisoning.

### Accurately reflect YJS's API

While users of YJS (or at least this one) want types to help us avoid accidentally corrupting our document schema, it's also important to recognize that YJS's low-level API does not include the concept of a schema. Therefore, there must be a way for users to bypass this schema validation, for instance to migrate documents from one version of software to another, or to opt out of typed schemas completely.

Additionally, the design of YJS's API precludes the ability to guarantee that fields are not undefined. Because of this, we should still include `undefined` as a possible return value of `map.get`, even when the map is bound to a schema type.

Finally, YJS only allows certain types of values to be included in a `Y.Map`. The type definitions should prevent users from accidentally trying to store invalid, unserializable values in a `Y.Map` under any circumstance.

### Supporting existing use cases

It's also important to maintain the ability to define types similar to what can be used today. While there does not appear to be a way to make this change non-breaking without significantly increasing code complexity, the migration path for any users who are happily using the existing types is quite simple:

```ts

const map = doc.getMap<number>();
// becomes...
const map = doc.getMap<Record<string, number>>();
```

While this migration step is unfortunate, my opinion is that the post-migration types will more clearly communicate to TS developers the shape of the data contained in the map.

## Testing

`npm run lint` passes with no type errors. Unfortunately I've had to include some `any` types in the internal implementations to get there due to limitations of JSDoc types and to avoid getting into logic changes, but I've kept these as limited in scope as I can. Of note, these changes to `Y.Map`'s type interface caught a couple of issues in test cases that easily could have led to runtime errors outside of a test case.

I've also included a typescript file I used to validate the public type interface, which also helps demonstrates the semantics under various cases. This file is not required for the change to work, and can be removed from the PR if you don't think it will be useful in the future.

I am probably not the most experienced YJS user out there, so I would appreciate anyone willing to volunteer to help test these types under more complex scenarios :slightly_smiling_face:

### DIFF
--- a/src/types/YMap.js
+++ b/src/types/YMap.js
@@ -22,6 +22,18 @@ import * as iterator from 'lib0/iterator'
 /**
  * @template T
  * @typedef {Extract<keyof T, string>} StringKey<T>
+ *
+ * Like keyof, but guarantees the returned type is a subset of string.
+ * `keyof` will include number and Symbol even if the input type requires string keys.
+ */
+
+/**
+ * @template T
+ * @typedef {readonly {
+ *            [K in keyof T]: [K, T[K]];
+ *          }[StringKey<T>][]} EntriesOf<T>
+ *
+ * Converts an object schema into a readonly array containing valid key-value pairs.
  */
 
 /**
@@ -59,12 +71,12 @@ export class YMapEvent extends YEvent {
 export class YMap extends AbstractType {
   /**
    *
-   * @param {Iterable<readonly [string, any]>=} entries - an optional iterable to initialize the YMap
+   * @param {EntriesOf<MapType>=} entries - an optional iterable to initialize the YMap
    */
   constructor (entries) {
     super()
     /**
-     * @type {Map<string,any>?}
+     * @type {Map<StringKey<MapType>, MapType[StringKey<MapType>]>?}
      * @private
      */
     this._prelimContent = null
@@ -88,9 +100,8 @@ export class YMap extends AbstractType {
    */
   _integrate (y, item) {
     super._integrate(y, item)
-    ;/** @type {Map<string, any>} */ (this._prelimContent).forEach((value, key) => {
-      // TODO: fix
-      this.set(/** @type {any} */ (key), value)
+    this._prelimContent?.forEach((value, key) => {
+      this.set(key, value)
     })
     this._prelimContent = null
   }
@@ -111,8 +122,7 @@ export class YMap extends AbstractType {
      */
     const map = new YMap()
     this.forEach((value, key) => {
-      // TODO: fix
-      map.set(key, /** @type {any} */ (value instanceof AbstractType ? /** @type {typeof value} */ (value.clone()) : value))
+      map.set(key, value instanceof AbstractType ? /** @type {typeof value} */ (value.clone()) : value)
     })
     return map
   }
@@ -185,7 +195,7 @@ export class YMap extends AbstractType {
   /**
    * Executes a provided function on once on every key-value pair.
    *
-   * @param {function(MapType[keyof MapType],StringKey<MapType>,YMap<MapType>):void} f A function to execute on every element of this YArray.
+   * @param {function(MapType[StringKey<MapType>],StringKey<MapType>,YMap<MapType>):void} f A function to execute on every element of this YArray.
    */
   forEach (f) {
     this._map.forEach((item, key) => {
@@ -207,7 +217,6 @@ export class YMap extends AbstractType {
   /**
    * Remove a specified element from this YMap.
    *
-   * TODO: type to only allow deletion of optional elements in MapType
    * @param {string} key The key of the element to remove.
    */
   delete (key) {
@@ -291,9 +300,3 @@ export class YMap extends AbstractType {
  * @function
  */
 export const readYMap = _decoder => new YMap()
-
-
-/** @type {YMap<{ foo: { hello: [3, 4, { hiThere: null }] }; }>} */
-const foo = new YMap();
-
-foo.has("")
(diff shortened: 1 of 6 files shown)


src/types/YMap.js
```diff
@@ -74,7 +99,7 @@ export class YMap extends AbstractType {
    */
   _integrate (y, item) {
     super._integrate(y, item)
-    ;/** @type {Map<string, any>} */ (this._prelimContent).forEach((value, key) => {
+    this._prelimContent?.forEach((value, key) => {
```
Smona, line 102: note this logic change, but if `_prelimContent` was null previously then this would have created a runtime error.

dmonad:
Thanks for the PR!

This is on the roadmap for v14, as it is a breaking change.

However, this feature is not "enough" for a new v14 release. I don't want to create too many releases for Yjs.

I'll keep this  PR around for now.

Note: I dislike the `.ts` file (this could be done in JavaScript and included in the tests).

Smona:
Feel free to point me towards the v14 roadmap (couldn't find it with a cursory search) -- I'd be happy to assist with any other (potentially breaking) type improvements you would like to include. These are just the main type issues I'm aware of.

In the meantime, I'll try to get the TS file migrated to JS and included in `npm run lint` so this is ready when you are.

julianpoy:
This issue is fairly impactful for TypeScript users. At the moment, I'm having to cast everything to yjs-types (https://www.npmjs.com/package/yjs-types) which aren't perfect.

manstie:
can we chuck in a fix for Y.Array's `.get` method returning `T` when it could be `T | undefined` 👉👈

robcresswell:
@dmonad Is there any more recent plans to merge this? Its still a pretty unpleasant experience for TS users, you need to write a bunch of custom typing. Why is there a reluctance to release a v14? Probably easier to have more releases with smaller breaking changes than one big release with many, I'd have thought.

If I can help with the PR somehow, please let me know, I'm a big yjs fan

djrenren:
Hey I just wanted to check on this change. It's still very difficult to get typing right with Y Maps. While this would be a breaking change (adopting it wouldn't type check for existing users), it would still be fully compatible at runtime meaning it can be safely incrementally adopted.

rcrogers:
Working types is a release-worthy feature!

Smona:
Okay, I think I've resolved all the outstanding requests! Apologies for the long delay. Latest `main` is merged in, and types & tests are both passing.

- Replaced the `.ts` file with JSDoc type tests. These will now cause type errors in `lint`/`dist` if they're failing, and make use of `@ts-expect-error` to confirm validations are working.
- Improved correctness of the return type of `YMap.toJSON()` by accounting for recursive calls on directly nested `AbstractTypes`. Also gave `YArray.toJSON()` the same treatment, so it returns a more constrained type if its type parameter has been explicitly set.
- Changed the return type of `YArray.get()` to `T | undefined`.
- Renamed `MapValue` to `SerializableValue` and constrained the `YArray` type parameter with it as well.
- Added default type parameters for `Y.Array` & `Y.Map` types (purely ergonomic).

@dmonad is there anything else I can do to help get this ready to merge? Any other type changes you would like to bundle in before cutting a release with these changes?

DawidWraga:
Would absolutely love to see this released soon! thanks.

dmonad:
Yjs @v14 is on the way! (see `diffing` branch)
It will take a couple of months, but it will include a bunch of ergonomic improvements.

julianpoy:
Hi @dmonad! Just wanted to check that this is part of the new v14 beta releases?

PR #618, opened 2024-02-29
Facilitate referencing UndoManager StackItem inside Type observers

### LINKED ISSUE #617: Ability to reference UndoManager StackItem inside Type observers
**Checklist**

 - [x] Are you reporting a bug? Use github issues for bug reports and feature requests. For general questions, please use https://discuss.yjs.dev/
- [x] Try to report your issue in the correct repository. Yjs consists of many modules. When in doubt, report it to https://github.com/yjs/yjs/issues/

**Is your feature request related to a problem? Please describe.**

I'd like to be able to reference the "in-flight" `StackItem` from an `AbstractType.observe` callback.

Consider the following: 

```javascript
import { Doc, UndoManager } from 'yjs'

const doc = new Doc()
const text = doc.getText('text')
const undoManager = new UndoManager([text])

text.observe((event) => {
    const origin = event.transaction.origin
    if(origin === undoManager) {
        console.log({
            undoLength: origin.undoStack.length, // 0
            redoLength: origin.redoStack.length  // 0
        })
    }
})

text.insert(0, 'abc')
undoManager.undo()
```

In this callback the `UndoManager` is in an intermediate state, where it has:
 - popped an item off the relevant stack, keeping the only reference to the `StackItem` in local scope
 - successfully applied the transaction, triggering the observer callback
 - has **not** yet torn down the `undoing` state and added the `StackItem` back to the relevant stack

**Describe the solution you'd like**

I'd like to push some document state and `StackItem` meta data to an external state store within the same procedure. It looks like the simplest way of doing this in the current architecture would be to add a temporary reference to the `StackItem` on the `UndoManager` while it is processing the item.

**Describe alternatives you've considered**

Currently the API only supports the following, requiring two separate change events in my state store. This causes unnecessary recalculation and also pushes the problem further downstream with additional re-rendering in my view-layer.

```javascript
undoManager.on('stack-item-popped', ({ type, stackItem }) => {
    // push some StackItem meta data to store
})

text.observe(event => {
    // push some event.changes data to store 
})
```

**Additional context**

 - Proposal #618

mylesj:
- Resolves #617

### DIFF
--- a/src/utils/UndoManager.js
+++ b/src/utils/UndoManager.js
@@ -191,6 +191,12 @@ export class UndoManager extends Observable {
      */
     this.undoing = false
     this.redoing = false
+    /**
+     * The currently popped stack item if UndoManager.undoing or UndoManager.redoing
+     *
+     * @type {StackItem|null}
+     */
+    this.doingStackItem = null
     this.lastChange = 0
     this.ignoreRemoteMapChanges = ignoreRemoteMapChanges
     this.captureTimeout = captureTimeout
@@ -331,10 +337,12 @@ export class UndoManager extends Observable {
    */
   undo () {
     this.undoing = true
+    this.doingStackItem = array.last(this.undoStack) ?? null
     let res
     try {
       res = popStackItem(this, this.undoStack, 'undo')
     } finally {
+      this.doingStackItem = null
       this.undoing = false
     }
     return res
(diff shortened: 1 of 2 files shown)


dmonad:
Hi @mylesj ,

thanks for the PR! I added my own spin on things and published it in `Yjs@13.6.13`. The variable is now named `um.currStackItem`.

PR #626, opened 2024-04-09
Update Readme who-is-using (thinkdeli.com)

satyajeetjadhav:
I wanted to create a really fast and simple writing app that worked offline and across devices. yjs is the key that made this possible. 
Thanks!


dmonad:
Sweet, thanks for sharing!

PR #627, opened 2024-04-11
update search marker count in INTERNALS.md

synix:
The `markPosition()` and `findMarker()` functions in `src/types/AbstractType.js` both handle **search marker** with an upper bound `maxSearchMarker`, which is defined as `const maxSearchMarker = 80`. but according to **INTERNALS.md**, it says `Yjs stores a cache of the 10 most recently looked up insert positions in the document.` Are they same concept? 
Give a try to update the value in **INTERNALS.md** if they are.


dmonad:
Yes, I changed it to 80 some time ago.

Thanks for improving the docs!

PR #629, opened 2024-04-13
remove outdated Y instance in comments

synix:
The `Y` class was replaced with `Y.Doc` some time ago, but some code comments is still related to `Y` instance or its `define` method, So try to update the outdated comments.


dmonad:
Thanks!

PR #641, opened 2024-05-02
missing state vector should be usable for update encoding

Horusiath:
The following test confirms an edge case when missing state vector is not correctly computed.

When we get an out-of-order update for a client, that has never been seen before, a computed missing state vector starts from that update start clock instead of 0.

Second commit contains proposed solution.

<sub><a href="https://huly.app/guest/yjs?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NjMzMmVkYTAyYzdiMjYwNDYzMmRiNmIiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctZ2l0aHViZG1vbmFkLXlqcy02NjBkNTc3Mi03MGYwNmEwYTIyLWQyOGNmNSIsInByb2R1Y3RJZCI6IiJ9.tCvMYXI_VVOmBqlgObtIjMlAXkYN97HEbppHz9EgQMQ">Huly&reg;: <b>YJS-816</b></a></sub>

### DIFF
--- a/src/utils/encoding.js
+++ b/src/utils/encoding.js
@@ -265,7 +265,10 @@ const integrateStructs = (transaction, store, clientsStructRefs) => {
    */
   const updateMissingSv = (client, clock) => {
     const mclock = missingSV.get(client)
-    if (mclock == null || mclock > clock) {
+    if (mclock === undefined) {
+      // we never seen that client before
+      missingSV.set(client, 0)
+    } else if (mclock > clock) {
       missingSV.set(client, clock)
     }
   }
--- a/tests/doc.tests.js
+++ b/tests/doc.tests.js
@@ -327,3 +327,45 @@ export const testSyncDocsEvent = async _tc => {
   t.assert(!ydoc.isSynced)
   t.assert(ydoc.whenSynced !== oldWhenSynced)
 }
+
+/**
+ * @param {t.TestCase} tc
+ */
+export const testMissingStateVector = tc => {
+  const d1 = new Y.Doc()
+  d1.clientID = 1
+  const m1 = d1.getMap('test')
+  /** @type {any[]} */
+  const updates = []
+  d1.on('update', (update) => {
+    updates.push(update)
+  })
+  m1.set('a', '1')
+  m1.set('b', '2')
+  m1.set('c', '3')
+  m1.set('d', '4')
+  m1.set('e', '5')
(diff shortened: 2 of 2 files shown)


dmonad:
I think the current implementation is preferable. The additional case is not needed.

When we know that operation  `o1` depends on the missing `o2(clock:3, client: 2)`, then we populate `missingSV.set(2, 3)`. Once we receive `o2`, we might figure out that we also need `o3(clock:2, client: 2)`. 

We don't need any more information to determine that something is missing. 

The reason why I think the current approach is preferable is as follows: We should not try to apply `o2 with missingSV( {2: 3} )` once we receive `o4(client: 2, clock 0)`, because it still does not resolve the missing dependency.

It is not required that `missingSV`  is "complete" (i.e. it contains all operations that the update might require). This would also require much more computation. But it should be sound (contain only operations that are required)

I think I get what you are trying to achieve. If you notice that there are missing operations, then you want to send the `missingSV` instead of recalculating the full state vector. There are other edge cases when this might not work in the current codebase of Yjs. One example: we only compute the first "missing operation" for each operation (first we report `origin`, then we might report `rightOrigin`). 

Furthermore, you still might have other lost operations that are no dependencies of other operations. Loosing updates in a client-server model with a reliable protocol (TCP/WS) is a REALLY bad thing. That means that somewhere in the codebase, updates are not propagated correctly. You can't detect all kinds of dataloss with `missingSV` - this only captures a small number of edge cases.

A quickfix would be to do a full resync in regular intervals. This would at least capture all kinds of dataloss. However, it might make sense to reevaluate the codebase itself. Often, usercode throws exceptions, which prevents the propagation of otherwise well-formed updates. It the case of an exception, the connection should be closed.

Horusiath:
Yeah, I think I've galloped too far. With this PR approach missing state vector would be actually equivalent to document's own state vector. Question would be: what things existing state vector could be used for?

dmonad:
The use-case of `missingSV` is only "if any of the ops in `missingSV` is applied, then we can retry applying the associated update". 

There is probably little overlap with the state vector which describes "what we have applied to our own document".

PR #648, opened 2024-06-06
Export mergeDeleteSets

fuadsaud:
Useful for comparing snapshots.

I tried reimplementing it in my own code but it depends on the `DeleteSet` class, which is also not exported.

<sub><a href="https://huly.app/guest/yjs?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NjYxOWZmNjhmODE1YzM1ZjQ0ZDdhMDAiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctZ2l0aHViZG1vbmFkLXlqcy02NjBkNTc3Mi03MGYwNmEwYTIyLWQyOGNmNSIsInByb2R1Y3RJZCI6IiJ9.BchwzO54K5QG9nzZJxSnFiQQn2D2EuEU2cHH2Ou21LI">Huly&reg;: <b>YJS-824</b></a></sub>

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -100,6 +100,7 @@ export {
   UpdateDecoderV1,
   UpdateDecoderV2,
   equalDeleteSets,
+  mergeDeleteSets,
   snapshotContainsUpdate
 } from './internals.js'
 


fuadsaud:
By the way: is there a good reason why the internals are not exported in the NPM module? Perhaps they could all be exported under an "internals" namespace to communicate that they're not part of the public api but still allow them to be used when needed.

fuadsaud:
@dmonad coming back to this one cause I found myself in a situation where I needed to get some internal types, particularly `DocEvents`, as I'm building some wrappers around the yjs API.

Do you think it would be possible to expose those somehow? I'm not familiar with the build process of yjs and how it exports the types from JSDoc.

Issue #638, opened 2024-04-26
Inconsistency in absolute position index across clients after an undo operation


nimeshnayaju:
**Checklist**

* [x] Are you reporting a bug? Use github issues for bug reports and feature requests. For general questions, please use https://discuss.yjs.dev/
* [x] Try to report your issue in the correct repository. Yjs consists of many modules. When in doubt, report it to https://github.com/yjs/yjs/issues/

**Describe the bug**
Absolute position index after an undo operation becomes inconsistent across different clients in a collaborative text editing session. I am mostly curios if this expected behavior. If so, this would make use of `RelativePosition` for commenting in a collaborative text document difficult to implement or almost unusable.

**To Reproduce**

See reproduction here: [Stackblitz Link](https://stackblitz.com/edit/vitest-dev-vitest-vc9ngf?file=test%2Fbasic.test.ts,test%2Fyjs.test.ts%3AL11-L11&initialPath=__vitest__/)

Steps:
1. Two clients (Client A and Client B) start with synchronized documents with a shared text type under key `text`. Changes made in Client A are applied to Client B.
2. Client A inserts `abcde` to the shared `text`.
3. Create a relative position at index 2 of the shared `text`.
4. Convert this relative position to absolute positions on both Y.js Doc on Client A and Client B, and verify that they point to the same position.
5. Delete the character at index 2 (`c`) from the shared `text` and verify that the changes (deletion) are reflected on both clients.
7. Perform an undo operation using UndoManager to revert the deletion.
8. Convert the same initial relative position again to absolute positions on both Client A and B.

**Expected behavior**
After undoing the deletion, both documents should revert to 'abcde' and the absolute position at index 4 should be consistent across both clients.

**Actual behavior**
The text reverts correctly, but the absolute position points to index 2 in Client A and incorrectly to index 3 in Client B.

**Environment Information**
 - Yjs v3.6.14

dmonad:
Yes, this is the expected behavior. I'll try to explain what is happening internally so you can understand it better. Below is a fix to get consistent behavior.

In Yjs we can't undo deletions (as in undeleting the internal item representation). Instead, we re-insert content. The advantage is that we can perform very aggressive garbage-collection. This means that after performing undo/redo, relative positions are not always placed at the position where they intuitively should be. This is a tradeoff - but I think it's worth it.

Intuitively, the relative position should be at index 2 after performing undo. We can't ensure that.

[...]

PR #655, opened 2024-07-09
Include `Item.redone` in binary encoding

markcrispo:
**Related issue:**
- https://github.com/yjs/yjs/issues/638

**Description:**
I'm looking to build a commenting solution using relative positions. I ran into the same issue described in #638, but in my case I would like all clients to be able to "follow the redo-path" as described by @dmonad [here](https://github.com/yjs/yjs/issues/638#issuecomment-2080220529), and looked into persisting into persisting `Item.redone` in the binary encoding of the document. 

**Implementation:**
In the `info` byte, I took `BIT5` away from `contentRefs` and used it to signal the presence of `redone`. We were using 5 bits to store the content type, but 4 bits is sufficient for the 10 possible options.

I updated the V1 and V2 Encoders and Decoders to write the `redone` ID if it exists, based on the implementations for `leftID` and `rightID`.

**Testing:**
I updated the second-last assertion in `testRelativePositionWithUndo`, which was testing for the opposite behaviour from what I was looking for. Now the test shows that both the local client, who originated the Undo operation, and the remote client who received it, both have the option whether or not to follow the redone path when resolving a relative position.

Please let me know if you have any feedback @dmonad 🙏 I'm not sure if this PR goes contrary to an intentional design decision!

### DIFF
--- a/src/structs/Item.js
+++ b/src/structs/Item.js
@@ -93,6 +93,7 @@ export const splitItem = (transaction, leftItem, diff) => {
     leftItem.rightOrigin,
     leftItem.parent,
     leftItem.parentSub,
+    leftItem.redone !== null ? createID(leftItem.redone.client, leftItem.redone.clock + diff) : null,
     leftItem.content.splice(diff)
   )
   if (leftItem.deleted) {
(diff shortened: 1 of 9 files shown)


dmonad:
I really didn't intend to keep this kind of information _forever_.

The undo manager "keeps" items that are referenced by the local undomanager. For your approach to work, you'd also need to disable gc for all items that were ever touched by an undomanager - essentially disabling gc for all users. Further tests will reveal that flaw.

This can certainly be done, but I wouldn't recommend it. Please consider the solution that I mentioned in the ticket. Another solution would be to make comments a part of the document.

PR #660, opened 2024-07-30
Add missing functions and remove erroneus function from API docs

mtreinik:
Documented `Y.Array.from()` and `yarray.clone()` that exist in the implementation.

There is no `ymap.get(index:number)`


mtreinik:
See also PR https://github.com/yjs/docs/pull/57

dmonad:
Thank you, much appreciated!

PR #664, opened 2024-08-29
Add QDAcity to `README.md`

jul13579:
Since we've contributed to `yjs` and use it in our product, I've wondered if we can enlist ourselves in the list of yjs users?

Thank you all so much :blush:


dmonad:
I'm happy to learn about new users :)

PR #665, opened 2024-09-05
add ScienHub as a user.

batchor:
Add <https://scienhub.com> to the user list.


dmonad:
Looks awesome, thanks for sharing!

PR #668, opened 2024-10-17
Return unobserve functions from observe and observeDeep methods

### LINKED ISSUE #327: Return and unobserve function from observe and observeDeep
I suggest having the observe and observeDeep functions return an `unobserve function` instead of `void`.  This follows common conventions and makes code cleaner. It will not break existing code either.

# Currrent Way
```js
let observer
yVal.observe( observer = () => {
  // do something
})
onUnmount(()=>{
 yVal.unobserve(observer)
})
```

# Suggested Way
```js
const unobserve = yVal.observe( () => {
  // do something
})
onUnmount(()=>{
 unobserve()
})
```

giraugh:
Closes #327

Went looking for this feature and spotted the issue so thought I'd try my hand at the simple change. Happy to make further changes or add/update tests if desired 👍 

Thanks for all your work on y.js :)

### DIFF
--- a/src/types/AbstractType.js
+++ b/src/types/AbstractType.js
@@ -364,20 +364,26 @@ export class AbstractType {
 
   /**
    * Observe all events that are created on this type.
+   * Returns a function to stop sending events to the provided callback once called
    *
    * @param {function(EventType, Transaction):void} f Observer function
+   * @returns {function():void}
    */
   observe (f) {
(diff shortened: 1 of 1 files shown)


dmonad:
Thanks for the PR!
I'm not sure if I like the approach. I'd prefer `observe(f)` to return `f`. I added a comment in #327.

PR #676, opened 2024-11-27
docs: remove `@toeverything/y-indexeddb`

himself65:
Upstream has removed this package in their repo, https://github.com/toeverything/AFFiNE/pull/6728


dmonad:
Too bad. But thanks for the update!

PR #681, opened 2024-12-16
fix: allow numbers in attribute values

### LINKED ISSUE #680: YXmlElement.clone() only copies string attribute values / does not support numbers
**Describe the bug**
`YXmlElement.clone()` has a check on line 97 https://github.com/yjs/yjs/blob/main/src/types/YXmlElement.js#L97 to verify whether the attribute value is a string and skips the attribute if the value isn't a string.

However, there are lots of use cases where the attribute value can be a number. The primary case we're running into is where a Prosemirror / YJS integration has paragraphs with indent values (and heading levels etc) that are numbers.

**To Reproduce**
Attempt to clone any `YXmlElement` that has non-string attribute values.

**Expected behavior**
Attribute values that are numbers should be copied over to the clone and not dropped.

**Environment Information**
YJS `main` / v13.6.20

tomi-bigpi:
Changes YXMLElement to allow numbers in attribute values

This seems like a common use case when working with YJS and Prosemirror/TipTap schemas, since anything from list values, indent values, to heading levels will contain numeric values instead of strings when converting between YJS and Prosemirror.

fixes #680

### DIFF
--- a/src/types/YXmlElement.js
+++ b/src/types/YXmlElement.js
@@ -25,7 +25,7 @@ import {
  * * An YXmlElement has attributes (key value pairs)
  * * An YXmlElement has childElements that must inherit from YXmlElement
  *
- * @template {{ [key: string]: ValueTypes }} [KV={ [key: string]: string }]
+ * @template {{ [key: string]: ValueTypes }} [KV={ [key: string]: string | number }]
  */
 export class YXmlElement extends YXmlFragment {
   constructor (nodeName = 'UNDEFINED') {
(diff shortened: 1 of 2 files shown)


hoangqwe159:
I do not think it is a good idea. It does not matter what value you get from Prosemirror or any editor, Yjs provides shared types and they should follow web standards. According to Web API, XML attribute values must always be quoted. That is why it should be always a string.

tomi-bigpi:
Well, MDN states for `setAttribute` that `Value: A string containing the value to assign to the attribute. Any non-string value specified is converted automatically into a string.` https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute That change is in the `toDOM` method (though only for `number`).

Additionally, note that this case came up in the first place because we're _cloning_ an existing element which already had somehow managed have a number as the value type. So one could just as easily argue that things should have already broken upstream when the doc/elements were created. At a minimum, there's a big difference in the behavior between creating elements and cloning them.

PR #688, opened 2025-01-09
security: add update size limits to prevent DOS attacks

### LINKED ISSUE #687: Denial of Service vulnerability: run-length encoding can be exploited to lock JS threads
## **Checklist**

### **Describe the bug**

It's possible to craft an invalid Uint8Array that will exploit run-length encoding to make Yjs' internals loop forever or eat up all your RAM when decoding an update.

## **To Reproduce**

Run following code snippets with nodejs and observe infinite loop or crashing.  Note that this list of reproduction cases might not be exhaustive.

You can also checkout [this PR](https://github.com/yjs/yjs/pull/688), comment out the diff to `update.js` and `encoding.js` and run the new test file to reproduce.

### Y.mergeUpdatesV2

This will abuse Content Length encoding by creating a big decoder, we could enforce a hard limit to decoder generator to fix it (cf [code](https://github.com/yjs/yjs/blob/main/src/utils/updates.js#L106)):

```ts
const buf = Buffer.from('00010000000001000100010200001602000001fdffffffff7f0000', 'hex');
const update = Y.encodeStateAsUpdateV2(new Y.Doc());
 // run forever
Y.mergeUpdatesV2([update, buf]);
```

This will create a huge number of struct (cf [code](https://github.com/yjs/yjs/blob/main/src/utils/updates.js#L46)) with a large number of state updates ([code](https://github.com/yjs/yjs/blob/main/src/utils/updates.js#L44)), introducing quadratic complexity.

```ts
const buf = Buffer.from('e4950000011800000100010000010000ed01be82e3c31c0102e49500000118000001180000010001ed00', 'hex');

const update = Y.encodeStateAsUpdateV2(new Y.Doc());
 // run forever
Y.mergeUpdatesV2([update, buf]);
```

This will loop forever in the While True loop on this statement ([code](https://github.com/yjs/yjs/blob/72205a688f266cc80930907972c8f7aa11be76e5/src/utils/updates.js#L426))

```ts
const buf = Buffer.from('8f018000000001aa010001020000160200e5c44314e7a68b93aeb5fd5de8269a8a590031d50f120130000000', 'hex');

const update = Y.encodeStateAsUpdateV2(new Y.Doc());
 // run forever
Y.mergeUpdatesV2([update, buf]);
```

### Y.mergeUpdates

```ts
const buf = Buffer.from('e49c100005ffff0500010000000101000000ed000000010100feb8c2e9ad87d91200000101ffedf6', 'hex');
const update = Y.encodeStateAsUpdate(new Y.Doc());
 // run forever
Y.mergeUpdates([update, buf]);
```

### Y.applyUpdateV2

```ts
const buf = Buffer.from('00012300000001028100001000c7dc00c47a800041abead60001000001000084000010ffc7dcff000000', 'hex');
// FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
Y.applyUpdateV2(new Y.Doc(), buf);
```

## **Expected behavior**

Instead of starting an infinite loop, Yjs should reject these invalid updates and throw an error.

## **Environment Information**
 - Node v23
 - Yjs 13.6.21

## Additional context

This issue was initially discovered by @adrienperonnetDT by fuzzy testing Yjs using [jsfuzz](https://gitlab.com/gitlab-org/security-products/analyzers/fuzzers/jsfuzz).

It has been shared privately to maintainers of Yjs and Yrs back in May of 2024 and has been fixed in Y.rs:
* https://github.com/y-crdt/y-crdt/issues/450 
* https://github.com/y-crdt/y-crdt/pull/455

gvergnaud:
Resolves https://github.com/yjs/yjs/issues/687

### DIFF
--- a/src/utils/limits.js
+++ b/src/utils/limits.js
@@ -1,0 +1,48 @@
+export const MAX_STRUCTS = 100_000
+export const MAX_UPDATES = 100_000
+export const MAX_GC_LENGTH = 100_000
+export const MAX_SKIP_LENGTH = 100_000
+
+/**
+ * @param {number} numOfStateUpdates
+ */
+export function assertMaxUpdates (numOfStateUpdates) {
+  if (numOfStateUpdates > MAX_UPDATES) {
+    throw new Error(
+      `This update exceeds the maximum number of updates. ${numOfStateUpdates} > ${MAX_UPDATES}`
+    )
+  }
+}
+
+/**
+ * @param {number} numberOfStructs
+ */
+export function assertMaxStructs (numberOfStructs) {
+  if (numberOfStructs > MAX_STRUCTS) {
+    throw new Error(
+      `This update exceeds the maximum number of structs. ${numberOfStructs} > ${MAX_STRUCTS}`
+    )
+  }
+}
+
+/**
+ * @param {number} len
+ */
+export function assertMaxSkipLength (len) {
+  if (len > MAX_SKIP_LENGTH) {
+    throw new Error(
+      `This skip length exceeds the limit. ${len} > ${MAX_SKIP_LENGTH}`
+    )
+  }
+}
+
+/**
+ * @param {number} len
+ */
+export function assertMaxGCLength (len) {
+  if (len > MAX_GC_LENGTH) {
+    throw new Error(
+      `This garbage collection update's length exceeds the limit. ${len} > ${MAX_GC_LENGTH}`
+    )
+  }
+}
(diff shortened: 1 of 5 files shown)


src/utils/limits.js
```diff
@@ -0,0 +1,49 @@
+export const MAX_STRUCTS = 100_000
+export const MAX_UPDATES = 100_000
+export const MAX_GC_LENGTH = 100_000
+export const MAX_SKIP_LENGTH = 100_000
```
gvergnaud, line 4: These limits are totally arbitrary. I honestly don't know what number would make sense here. 

Another approach I thought about was to increment a loop counter, and use the same limit for structs and updates. I don't think this would work for GC and skip item lengths, though.

PR #689, opened 2025-01-09
fix(yText): applyDelta should support both Delta and Ops[]

ykou-clickup:
Fixed an issue that the yText.applyDelta() accepted only Ops[], but not Delta.

### DIFF
--- a/src/types/YText.js
+++ b/src/types/YText.js
@@ -972,15 +972,16 @@ export class YText extends AbstractType {
     if (this.doc !== null) {
       transact(this.doc, transaction => {
         const currPos = new ItemTextListPosition(null, this._start, 0, new Map())
-        for (let i = 0; i < delta.length; i++) {
-          const op = delta[i]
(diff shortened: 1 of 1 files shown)


dmonad:
Makes sense. Thanks!

dmonad:
I'm sorry, I reverted the changes again. 

Generally, I avoid polymorphism in my libraries. The arguments must always be of the same type. It can't be a Quill Delta or an Array of ops.

In Yjs, I only use the array of operations.

I know, this is not the same as in Quill. I rather would like to go for a more minimalistic interface.

dmonad:
It's time that I create proper typings for the delta op. For now, I updated the types, so the method accepts an `Array<any>`. At least, that should give you a warning when you try to supply the Delta object.

ykou-clickup:
> It's time that I create proper typings for the delta op. For now, I updated the types, so the method accepts an `Array<any>`. At least, that should give you a warning when you try to supply the Delta object.

@dmonad you may want to update the README file as well 
image

PR #701, opened 2025-03-18
README: Remove duplicate btw mention

hacklschorsch:
'btw' was mentioned twice; remove one mention.


dmonad (approved):
Right, thank you!!

PR #702, opened 2025-03-25
Update README.md to add https://github.com/kapv89/k_yrs_go to persistence providers

kapv89:
Ref: https://x.com/kevin_jahns/status/1904252641124753641


dmonad:
Thanks for sharing!

PR #707, opened 2025-04-22
optimize readAndApplyDeleteSet performance on large deletes

catssay:
When calling applyUpdate with large text and many deletes, frequent structs.splice calls become a performance bottleneck.
To improve this, we replace the repeated splice operations by rebuilding the structs array from scratch — combining the existing and newly added structs — while preserving the same side effects as before.

### DIFF
--- a/src/utils/DeleteSet.js
+++ b/src/utils/DeleteSet.js
@@ -12,6 +12,7 @@ import * as math from 'lib0/math'
 import * as map from 'lib0/map'
 import * as encoding from 'lib0/encoding'
 import * as decoding from 'lib0/decoding'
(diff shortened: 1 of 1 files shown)


dmonad:
Can you please make sure that the tests run?

Also, do you have some numbers / test-cases  that confirm that this improves performance? Does this decrease performance in some cases?

catssay:
> Can you please make sure that the tests run?
> 
> Also, do you have some numbers / test-cases that confirm that this improves performance? Does this decrease performance in some cases?

Sure, I will make sure tests run and do some performance testing when I get some free time soon.

PR #709, opened 2025-05-07
Diffing | v14

dmonad:


attributing-content.md
```diff
@@ -0,0 +1,121 @@
...
+// we render the attributed content with the attributionManager
+let attributedContent = ytext.getContent(attributionManager)
+console.log(JSON.stringify(attributedContent.toJSON().ops, null, 2))
+let expectedContent = delta.create().insert('Hell', { italic: true }, { attributes: { italic: ['Bob'] } }).insert('o ').insert('World', {}, { delete: ['Bob'] }).insert('attributions', {}, { insert: ['Bob'] }).insert('!')
+t.assert(attributedContent.equals(expectedContent))
+
+// this is how the output would look like
+const output = [
+  {
+    "insert": "Hell",
```
YousefED, line 66: I find `insert` a bit confusing, as it's always "insert" even if we're not talking about an _insertion_, and we already use the term `insert` in `attribution` as well.

Some ideas:
- use `content` instead of `insert`?
- alternatively, use the terms from the Quill change format (`retain`, `delete` and `insert`), but I'm not sure that's compatible with what you want to represent?

attributing-content.md
```diff
@@ -0,0 +1,121 @@
...
+  {
+    "insert": "World",
+    "attribution": {                // the insertion "World" was deleted by Bob
+      "delete": [
+        "Bob"
+      ]
+    }
+  },
+  {
+    "insert": "attributions",       // the insertion "attributions" was inserted by Bob
```
YousefED, line 90: let's not use "attributions" in demo text because it's going to make the example code / json more confusing 😅

YousefED (commented):
I haven’t had a chance to dive deep into the code ye (@nickthesick will start playing around with it first), but I did take a quick look at the example provided. I added some minor feedback on terminology while I still have that "first sight / rookie" perspective

attributing-content.md
```diff
@@ -0,0 +1,121 @@
...
+In order to implement a Google Docs-like versioning feature, we want to be able
+to attribute content with additional information (who created the change,
+when was this change created, ..).
+
+When we click on a version in Google Docs, we might get annotated changes like
+this:
+
+```
+# E.g. If Bob appends "world" to the previous version "hello "
+[{ insert: 'hello' }, { insert: 'world', color: 'blue', creator: 'Bob', when: 'yesterday' }]
```
YousefED, line 23: (see my other comment re. `insert`)

attributing-content.md
dmonad: I appreciate the feedback!

The operations are Quill Deltas, actually. They are used to represent the full state (in this case, the ops will be `insert` operations only). But they can also be used to represent actual changes (in which case we would use `retain` and `delete` as well).

Two things I've been thinking about:
`attribution` could be named `annotation` - which makes it less similar to `attributes`.
`attribution.insert` could be named `attribution.insertBy`. Do you think this might improve clarity?

attributing-content.md
dmonad: Fair point ^^

attributing-content.md
YousefED: I think those would be two valuable improvements indeed and would add more clarity!

One q about your proposal:

> attribution.insert could be named attribution.insertBy. Do you think this might improve clarity?

are these attributions (annotations) always about an `author`? Or can it also contain a timestamp, or something else? In that case `insertBy` wouldn't always make sense, right?

Let's continue the discussion a little bit so we're sure to land on the most sensible solution:

- I can see how you're modeling them after quill `insert` deltas. However, because we also use it to represent _deleted_ content our use case is a little bit different, right? In Quill, a document can only consist of `insert` operations and plainly concatenating everything would _always_ resolve in the document. In our case, we'd need to filter out content annotated as `deleted` to get to the "actual" document
- Quill also has a way to represent changes (with `retain` and `delete`), but I also don't think this matches one-on-one because `getContent` doesn't (and shouldn't afaik) return a diff
- In the current solution; if I don't pass an AnnotationManager, I suppose there will still be "deleted" annotations for deleted content, right (otherwise there's no way to figure out something has been deleted after calling `getContent`)? Does that make sense or will this be confusing?

Maybe it's a minor detail - I think what you propose (annotations and `insertBy`) is already a great improvement, but I figured worth to give the above some extra thought

PR #731, opened 2025-08-18
Add doc.beginTransaction() → callback, as an alternative to doc.transact()

percevalw:
Hello,

This PR introduces doc.beginTransaction() which start a transaction and return a finisher callback. It lets callers wrap transaction code with try/finally instead of passing a function to transact.

This is useful when a callback isn’t feasible, e.g., integrating with code that expects begin/end semantic, so the transaction can span arbitrary control flow and multiple functions.

```js
const ydoc = new Y.Doc()
const yxml = ydoc.getXmlFragment('')

const endTransaction = ydoc.beginTransaction('bulk')
try {
  for (let i = 0; i < 10; i++) yxml.push([new Y.XmlText('a')])
} finally {
  endTransaction()
}
```

### DIFF
--- a/src/utils/Doc.js
+++ b/src/utils/Doc.js
@@ -11,6 +11,7 @@ import {
   YXmlElement,
   YXmlFragment,
   transact,
+  beginTransaction,
   applyUpdate,
   ContentDoc, Item, Transaction, YEvent, // eslint-disable-line
   encodeStateAsUpdate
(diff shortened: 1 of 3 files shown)


dmonad:
It's usually better to create an issue first, so you can avoid the extra work.

First, I would need a good reason to implement a feature. I don't like to implement the same thing twice. I understand that some people would prefer `begin`/`end` transactions, like some database frameworks implement them. However, I prefer the more restrictive callback approach because it prevents a bunch of pitfalls.

- Users will definitely forget to call `endTransaction`, which will lead to issues.
- How would you handle creating two transactions at the same time? 

It's very unlikely that I will merge this PR without a good reason. If there was a reason, I would require that we only implement a single method to create transactions.

Issue #745, opened 2025-11-08
Tests failing in node.js, also on CI


raimohanska:
More than 90 tests fail with errors like

```
[87/230] array: repeat generating yarray tests43
  sync protocol doesnt support v2 protocol yet, fallback to v1 encoding
  TypeError: leftStruct.splice is not a function
```

In the output, there is this warning `Yjs was already imported. This breaks constructor checks and will lead to issues!` which might suggest that this happens because Y.js is imported twice. Which in turn might be because the tests import y-protocols that has Y.js as dependency.

I ran `git bisect` for this and it looks like the first broken commit is

```
584ad5fc9a330b659ed965b6d391234deb2240b9 is the first bad commit
commit 584ad5fc9a330b659ed965b6d391234deb2240b9 (HEAD)
Author: Kevin Jahns <kevin.jahns@protonmail.com>
Date:   Tue Oct 28 17:51:10 2025 +0100

    bump deps and use new delta format
```

More precisely it's the "bump deps" part that makes the tests fail. If I checkout the package.json and package-lock.json from the previous commit, only 10 tests fail at this commit, for different reasons (not delving into that).

If I remove package-lock.json, the tests start failing for older versions as well.

I didn't have time to dig deeper, but might do later. It would be great to have the tests fixed.

As a side-note, I might consider stripping down some of the test output as well - they are very verbose :)

PR #746, opened 2025-11-09
Fix tests by using an ES module loader when running tests

raimohanska:
Because of transitive dependencies, a Y.js dependency is installed under node_modules and is imported by @y/protocols package that's used in some Y.js tests. This import causes two separate Y.js instances to be loaded, which in turn wreaks havoc, as a lot of `instanceof` or other checks in the Y.js core code start to fail.

This Pull Request solves the problem by adding a `--loader` parameter to Node.js when running tests. The given loader script always loads `./src/index.js` instead of anything under `node_modules` when the package `yjs` is loaded by any of our devDependencies.

This solves #745. I have tested this locally with Node.js versions 16 and 24.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   },
   "scripts": {
     "clean": "rm -rf dist",
-    "test": "NODE_ENV=development node ./tests/index.js --repetition-time 50",
+    "test": "NODE_ENV=development node --loader ./tests/yjs-alias-loader.js ./tests/index.js --repetition-time 50",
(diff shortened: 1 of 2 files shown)


raimohanska:
Just checking: would you like to merge this? I mean, it would be great to get some green on the Actions tab, right? 😄 I know the approach is not optimal - maybe the best way would be to somehow fix the dependencies to avoid this "two y.js instances" situation, but the loader approach here seems to get us to green builds for now.

PR #753, opened 2025-12-17
Support attributing content in `DiffAttributionManager` & `SnapshotAttributionManager`

nperez0111:
This just adds an optional parameter to add your own default attribution for `DiffAttributionManager` & `SnapshotAttributionManager`

### DIFF
--- a/src/utils/AttributionManager.js
+++ b/src/utils/AttributionManager.js
@@ -9,7 +9,6 @@ import {
   insertIntoIdSet,
   diffIdMap,
   createIdMap,
-  createAttributionItem,
   mergeIdMaps,
   createID,
   mergeIdSets,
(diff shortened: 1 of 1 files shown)


src/utils/AttributionManager.js
```diff
@@ -553,17 +555,18 @@ export class SnapshotAttributionManager extends ObservableV2 {
   /**
    * @param {Snapshot} prevSnapshot
    * @param {Snapshot} nextSnapshot
+   * @param {Array<import('./IdMap.js').AttributionItem<any>>} [attrs]
    */
-  constructor (prevSnapshot, nextSnapshot) {
+  constructor (prevSnapshot, nextSnapshot, attrs = [createAttributionItem('change', '')]) {
```
dmonad: I know used the `change` attribution here before, but I think this should just be empty.

PR #758, opened 2025-12-24
[applyDelta] allow specifying origin and local for the `applyDelta` transaction

nperez0111:
I figured that it was useful for the `applyDelta` to allow specifying the origin for the transaction (so that I can filter by the origin, like I normally would).

I wasn't sure whether the `local` option needed to be configurable or not, but I included it anyway.

I also wasn't sure whether the attributionsManager should actually be under this new `opts` object. But, I didn't want to break the interface so I just added it as another param

### DIFF
--- a/src/types/AbstractType.js
+++ b/src/types/AbstractType.js
@@ -704,10 +704,13 @@ export class AbstractType {
    *
    * @param {delta.Delta<any,any,any,any,any>} d The changes to apply on this element.
    * @param {AbstractAttributionManager} am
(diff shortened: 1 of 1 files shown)


dmonad:
Hmm... not sure about this one. I still recommend doing `ydoc.transact(() => { type.applyDelta(..) }, origin)`.

You are right that applyDelta will become the preferred method to manipulate the types, so it makes sense to give it more responsibilities. But transact is still necessary, and I don't necessarily want to duplicate features right now.

PR #775, opened 2026-05-08
Fix undo of attr-change + delete corrupting remote sync (#757)

### LINKED ISSUE #757: UndoManager: Restored XmlText attributes missing on remote after setAttribute + delete + undo

**Checklist**

**Describe the bug**

When using `Y.UndoManager` to undo an attribute change followed by a deletion of an `XmlText` element, the restored element's attributes are not correctly encoded when syncing to remote clients.

Local state is correct, but when the document is encoded (via `encodeStateAsUpdate` or `encodeStateAsUpdateV2`) and applied to a remote document, some attributes are missing.

The bug occurs when ALL of these conditions are met:
1. An `XmlText` element exists with attributes (e.g., `{type: "button", test: true}`)
2. An attribute is modified via `setAttribute()` (e.g., `type: "button"` to `type: "paragraph"`)
3. The element is deleted
4. Both operations are undone together (same undo stack item)
5. The document state is encoded and sent to a remote client

The bug does NOT occur when:
 - Only attribute change + undo (no delete)
 - Only delete + undo (no attribute change)

**To Reproduce**

```js
import * as Y from 'yjs'

const doc = new Y.Doc()
const root = doc.get('sharedRoot', Y.XmlText)

// Create an XmlText element with two attributes
const button = new Y.XmlText()
button.setAttribute('type', 'button')
button.setAttribute('test', true)
button.insert(0, 'Click me')

root.insertEmbed(0, button)

const undoManager = new Y.UndoManager(root)
undoManager.stopCapturing()

// Step 1: Change an attribute
button.setAttribute('type', 'paragraph')

// Step 2: Delete the element
root.delete(0, 1)

// Step 3: Undo both operations
undoManager.undo()

// Local state is correct
console.log('Local:', root.toDelta()[0].insert.getAttributes())
// Output: { type: 'button', test: true }

// Remote state is wrong
const remoteDoc = new Y.Doc()
Y.applyUpdateV2(remoteDoc, Y.encodeStateAsUpdateV2(doc))
const remoteRoot = remoteDoc.get('sharedRoot', Y.XmlText)

console.log('Remote:', remoteRoot.toDelta()[0].insert.getAttributes())
// Output: { test: true } - "type" attribute is missing
```

**Expected behavior**
The remote document should have the same attributes as the local document after applying the encoded state: `{ type: 'button', test: true }`

**Environment Information**
- Node.js v20.19.10
- yjs v13.6.28 (and earlier versions, also tested on v13.5.41)

**Additional context**
This bug was discovered in a slate-yjs based collaborative editor where undo operations were causing state corruption on the server.

It happens when:
- user deletes a button with backspace, which converts the button to an empty paragraph, attribut change `type: "button"` -> `type: "paragraph"`)
- then they delete that empty paragraph, which deletes the element
- and then they undo those operations
- the state on the client is fine, but the state synced to the server (we use hocuspocus/server) doesn't have the `type: "button"` attribute

**Proposed Fix**

In `redoItem`, after determining left for map entries, check if `left.parent` matches the target `parentType`. If not, set `left = null` so the parent info gets encoded directly:

```js
// After line 224 in src/structs/Item.js
if (left !== null && left.parent !== parentType) {
  left = null
}
```

This ensures that when the restored attribute's left points to an item with a different parent (the old deleted element), we don't use it as a reference. Instead, the parent info will be encoded directly, and remote clients will correctly attach the attribute to the restored element.

I have a working fix with tests and can submit a PR if helpful.

* [ ] I'm a sponsor 💖
* [x] This issue is a blocker for my project.

ppiotrowicz:
## Summary

Fixes #757. When a `setAttribute` and a delete on the same `XmlText` are
captured in one undo stack item, undo restored the attribute correctly
locally but the encoded update sent to remote peers was missing the
restored attribute.

In `redoItem`, the redone attribute item ended up with a `left` pointing
under the original (now-deleted) parent. `Item.write` skips writing parent
info when `origin` is set, so the remote derived parent from `origin` and
attached the attribute to the wrong, deleted type.

The fix: after determining `left` for parentSub items, drop it if its
parent doesn't match the redo target — forcing the encoder to write
parent info directly.

## Test plan

- [x] Added regression test `testUndoSetAttributeAndDeleteSyncsAttributes`
      reproducing the exact scenario from the issue (local + remote
      assertions). Verified it fails on `v13` without the fix and passes
      with it.
- [x] Full test suite passes (208/208).

### DIFF
--- a/tests/undo-redo.tests.js
+++ b/tests/undo-redo.tests.js
@@ -811,3 +811,35 @@ export const testUndoDoingStackItem = async (_tc) => {
   t.compare(metaRedo, '42', 'currStackItem is accessible while redoing')
   t.compare(undoManager.currStackItem, null, 'currStackItem is null after observe/transaction')
 }
+
+/**
+ * @see https://github.com/yjs/yjs/issues/757
+ * @param {t.TestCase} _tc
+ */
+export const testUndoSetAttributeAndDeleteSyncsAttributes = _tc => {
+  const doc = new Y.Doc()
+  const root = /** @type {Y.XmlText} */ (doc.get('sharedRoot', Y.XmlText))
+  const button = new Y.XmlText()
+  button.setAttribute('type', 'button')
+  button.setAttribute('test', true)
+  button.insert(0, 'Click me')
+  root.insertEmbed(0, button)
+
+  const undoManager = new Y.UndoManager(root)
+  undoManager.stopCapturing()
+
+  button.setAttribute('type', 'paragraph')
+  root.delete(0, 1)
+  undoManager.undo()
+
+  const expectedAttrs = { type: 'button', test: true }
+  const expectedText = 'Click me'
+  t.compare(root.toDelta()[0].insert.getAttributes(), expectedAttrs)
+  t.compare(root.toDelta()[0].insert.toString(), expectedText)
+
+  const remoteDoc = new Y.Doc()
+  Y.applyUpdateV2(remoteDoc, Y.encodeStateAsUpdateV2(doc))
+  const remoteRoot = /** @type {Y.XmlText} */ (remoteDoc.get('sharedRoot', Y.XmlText))
+  t.compare(remoteRoot.toDelta()[0].insert.getAttributes(), expectedAttrs)
+  t.compare(remoteRoot.toDelta()[0].insert.toString(), expectedText)
+}
--- a/src/structs/Item.js
+++ b/src/structs/Item.js
@@ -223,6 +223,10 @@ export const redoItem = (transaction, item, redoitems, itemsToDelete, ignoreRemo
     } else {
       left = parentType._map.get(item.parentSub) || null
     }
+    // drop cross-parent left so origin doesn't mislead the remote (#757)
+    if (left !== null && left.parent !== parentType) {
+      left = null
+    }
   }
   const nextClock = getState(store, ownClientID)
   const nextId = createID(ownClientID, nextClock)


tests/undo-redo.tests.js
```diff
@@ -811,3 +811,35 @@ export const testUndoDoingStackItem = async (_tc) => {
...
+  undoManager.stopCapturing()
+
+  button.setAttribute('type', 'paragraph')
+  root.delete(0, 1)
+  undoManager.undo()
+
+  const expectedAttrs = { type: 'button', test: true }
+  const expectedText = 'Click me'
+  t.compare(root.toDelta()[0].insert.getAttributes(), expectedAttrs)
+  t.compare(root.toDelta()[0].insert.toString(), expectedText)
```
ppiotrowicz, line 838: Local version was correct

tests/undo-redo.tests.js
```diff
@@ -811,3 +811,35 @@ export const testUndoDoingStackItem = async (_tc) => {
...
+
+  const remoteDoc = new Y.Doc()
+  Y.applyUpdateV2(remoteDoc, Y.encodeStateAsUpdateV2(doc))
+  const remoteRoot = /** @type {Y.XmlText} */ (remoteDoc.get('sharedRoot', Y.XmlText))
+  t.compare(remoteRoot.toDelta()[0].insert.getAttributes(), expectedAttrs)
+  t.compare(remoteRoot.toDelta()[0].insert.toString(), expectedText)
```
ppiotrowicz, line 844: But remote had only the `test: true` attribute.

PR #778, opened 2026-05-17
fix stale JSDoc examples after v14 API rename

heylakatos:
## Summary

Two JSDoc `@example` blocks reference Doc/YMap methods that were removed in the v14 API rename:
- `src/utils/Snapshot.js` — `ydoc.getText()` → `ydoc.get()`
- `src/utils/Transaction.js` — `ydoc.getMap('map')` → `ydoc.get('map')`, `map.set(...)` → `map.setAttr(...)`

These examples no longer compile against the current API; anyone copying them from the docs would hit a runtime error.

## Test plan

- [x] `npm test` — all tests pass
- [x] Verified replacements against actual API:
    - `Doc#get(key='', name=null)` at `src/utils/Doc.js:197`
    - `YType#setAttr` at `src/ytype.js:1169`

### DIFF
--- a/src/utils/Snapshot.js
+++ b/src/utils/Snapshot.js
@@ -119,11 +119,11 @@ export const splitSnapshotAffectedStructs = (transaction, snapshot) => {
 /**
  * @example
  *  const ydoc = new Y.Doc({ gc: false })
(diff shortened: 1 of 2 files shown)


dmonad:
thanks!

PR #787, opened 2026-07-10
Don't delete nested Yjs types from `oldValue`

### LINKED ISSUE #786: Deleting non-empty Y.Map child from Y.Map triggers YMapEvent with **empty Y.Map** as `oldValue`.

**Checklist**

* [ ] Try to report your issue in the correct repository. Yjs consists of many modules. When in doubt, report it to https://github.com/yjs/yjs/issues/

**Describe the bug**
The `YMapEvent.oldValue` is an *empty* YMap when deleting a non-empty YMap child from a YMap parent it is nested in.

**To Reproduce**
1. Install YJS version 13.6.31
2. Write this file `test.mjs`:
```javascript
import * as Y from "yjs";

const ydoc = new Y.Doc()
const ymap = ydoc.getMap("map");

ymap.observe(ymapEvent => {
  // sample code.
  ymapEvent.changes.keys.forEach((change, key) => {
    if (change.action === 'add') {
      console.log(`ymap: Property "${key}" was added. Initial value: `, ymap.get(key).toJSON())
    } else if (change.action === 'update') {
      console.log(`ymap: Property "${key}" was updated. New value: "${ymap.get(key)}". Previous value: `,change.oldValue)
    } else if (change.action === 'delete') {
      console.log(`ymap: Property "${key}" was deleted. New value: undefined. Previous value: `,change.oldValue.toJSON())
    }
  })
})

const ymap2 = new Y.Map();

ymap2.observe(ymapEvent => {
  // sample code.
  ymapEvent.changes.keys.forEach((change, key) => {
    if (change.action === 'add') {
      console.log(`ymap2: Property "${key}" was added. Initial value: `, ymap.get(key).toJSON())
    } else if (change.action === 'update') {
      console.log(`ymap2: Property "${key}" was updated. New value: "${ymap.get(key)}". Previous value: `,change.oldValue)
    } else if (change.action === 'delete') {
      console.log(`ymap2: Property "${key}" was deleted. New value: undefined. Previous value: `,change.oldValue.toJSON())
    }
  })
})

ymap2.set('hello', 'world')

ymap.set('key', ymap2)
ymap.delete('key')
```
3. Run `node test.mjs` (but this bug also occurs on the web browser, with y-websocket and y-indexeddb attached)
4. See this in the console:
```javascript
ymap: Property "key" was added. Initial value:  { hello: 'world' }
ymap: Property "key" was deleted. New value: undefined. Previous value:  {}
```
(incorrect, since if "hello" were deleted from YMap first that would have appeared as a log.)

**Expected behavior**
```javascript
ymap: Property "key" was added. Initial value:  { hello: 'world' }
ymap: Property "key" was deleted. New value: undefined. Previous value:  { hello: 'world' }
```

**Screenshots**
Not applicable.

**Environment Information**
 - Browser / Node.js Node.js OR Chromium OR Firefox
 - Yjs version: 13.6.31

**Additional context**
I'm away this evening, but I'd be willing to look into fixing this bug tomorrow (Friday).

* [ ] I'm a sponsor 💖
* [x] This issue is a blocker for my project.

WebCoder49:
Firstly, I know this is a late time to send the bug, for Europeans. Don't feel pressured to review it too quickly or soon; my code likely contains bugs (though it passes all existing tests on `npm run test`, plus my new tests).

(TL;DR look at the code, the bullet points below "purpose", and my last few paragraphs.)

## Purpose
Fixes #786, letting y-shared-type  (`Y.Map`, `Y.Text`, and `Y.Array` have been tested) **children of `Y.Map` elements show up in `YMapEvent.changes.keys.get("name").oldValue`** during `observe` and `observeDeep` callbacks. Notable points include:
* No cloning occurs, so reference comparisons still work (this is in the test suite): `oldValue === ytext1` where `yext1` was defined as `const ytext1 = new Y.Text() ; map0.set('text', ytext1)`
* As I believe (but not have checked) occurred previously, taking the deleted `oldValue` or one of its child y-shared-types and, without cloning it, placing it elsewhere in the document does not work. That old value is marked as deleted (or for the case of child items will be deleted immediately after the observe callback, at the end of the transaction) so will be garbage-collected and not end up where you moved it in the long-term. This is easy to solve by cloning the y-type before placing it elsewhere in the document.
* Similarly, placing the `oldValue` in a variable during the observe callback without cloning it, and accessing the variable later outside the callback and after the transaction that deleted it, will give an empty type. This is similarly easy to solve by cloning the y-type in the observer function.

Would you be able to put these points in the documentation for version 13?

## Approach
My method does not change when the initial markDeletes happen; rather it temporarily removes the relevant delete marks while observer callbacks happen, and adds them back before the transaction ends and the garbage collector kicks in. I think the more drastic change of adjusting when initial markDeletes happen may prevent the need for the caveats listed above, but would also be a breaking change, and probably cause different unintuitive points.

Please let me know if you have any questions or requests about my code, tests, etc.

## Status
As of commit `a9052a4`, the tests pass (for `YMap`, `YText`, and `YArray` children of `YMap`s).

***I have not yet written any tests for xml-related children of `YMap`s, or any children of ~~`YArray` or~~ xml-related types, although this fix is general enough to probably affect them too. Should I write those tests?***

Thanks for your time!

### DIFF
--- a/src/structs/Item.js
+++ b/src/structs/Item.js
@@ -310,6 +310,13 @@ export class Item extends AbstractStruct {
      * @type {number} byte
      */
     this.info = this.content.isCountable() ? binary.BIT2 : 0
+
+    /**
+     * Has meaning only when this.deleted: True if this item was deleted only implicitly,
+     * due to the deletion of a map containing this item under a key; False in all other
+     * kinds of deletion.
+    */
+    this.deletedImplicitly = false;
   }
 
   /**
@@ -612,18 +619,27 @@ export class Item extends AbstractStruct {
    * Mark this Item as deleted.
    *
    * @param {Transaction} transaction
+   * @param {Boolean} implicitly True if this deletion was implicit,
+   * due to the deletion of a map containing this item under a key;
+   * False in all other kinds of deletion.
    */
-  delete (transaction) {
+  delete (transaction, implicitly=false) {
     if (!this.deleted) {
       const parent = /** @type {AbstractType<any>} */ (this.parent)
       // adjust the length of parent
       if (this.countable && this.parentSub === null) {
         parent._length -= this.length
       }
+      if(implicitly) {
+        this.deletedImplicitly = true;
+      }
       this.markDeleted()
       addToDeleteSet(transaction.deleteSet, this.id.client, this.id.clock, this.length)
       addChangedTypeToTransaction(transaction, parent, this.parentSub)
       this.content.delete(transaction)
+    } else if(!implicitly) {
+      // Previously deleted implicitly, but now deleted explicitly.
+      this.deletedImplicitly = false;
     }
   }
 
--- a/src/structs/Item.js
+++ b/src/structs/Item.js
@@ -313,8 +313,8 @@ export class Item extends AbstractStruct {
 
     /**
      * Has meaning only when this.deleted: True if this item was deleted only implicitly,
-     * due to the deletion of a map containing this item under a key; False in all other
-     * kinds of deletion.
+     * due to the deletion of a map containing this item under a key, AND the transaction
+     * for this implicit deletion has not yet happened; False in all other kinds of deletion.
     */
     this.deletedImplicitly = false;
   }
--- a/src/utils/Transaction.js
+++ b/src/utils/Transaction.js
@@ -4,6 +4,7 @@ import {
   writeDeleteSet,
   DeleteSet,
   sortAndMergeDeleteSet,
+  iterateDeletedStructs,
   getStateVector,
   findIndexSS,
   callEventHandlerListeners,
@@ -276,6 +277,22 @@ const cleanupTransactions = (transactionCleanups, i) => {
        * @type {Array<function():void>}
        */
       const fs = []
+
+      /**
+       * @type {Array<Item>}
+       */
+      const itemsToRemarkAsDeleted = [];
+      iterateDeletedStructs(transaction, ds, (item) => {
+        if(item instanceof Item && item.deletedImplicitly) {
+          // Unmark this implicitly-deleted item so it appears in the oldValue of YMapEvents.
+          item.deleted = false
+          item.deletedImplicitly = false // Allow it to be garbage collected when returned
+
+          // Implicitly-deleted items were still deleted so must be remarked before garbage collection!
+          itemsToRemarkAsDeleted.push(item)
+        }
+      })
+
       // observe events on changed types
       transaction.changed.forEach((subs, itemtype) =>
         fs.push(() => {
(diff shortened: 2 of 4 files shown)


src/structs/Item.js
```diff
@@ -310,6 +310,13 @@ export class Item extends AbstractStruct {
      * @type {number} byte
      */
     this.info = this.content.isCountable() ? binary.BIT2 : 0
+
+    /**
+     * Has meaning only when this.deleted: True if this item was deleted only implicitly,
+     * due to the deletion of a map containing this item under a key, AND the transaction
+     * for this implicit deletion has not yet happened; False in all other kinds of deletion.
+    */
+    this.deletedImplicitly = false;
```
dmonad, line 319: This makes sense, but we want to avoid additional properties on Item

src/utils/Transaction.js
```diff
@@ -264,6 +265,22 @@ const cleanupTransactions = (transactionCleanups, i) => {
     const store = doc.store
     const ds = transaction.deleteSet
     const mergeStructs = transaction._mergeStructs
+
+    /**
+     * @type {Array<Item>}
+     */
+    const itemsToReMarkAsDeleted = [];
+    iterateDeletedStructs(transaction, ds, (item) => {
```
dmonad, line 273: This would be too expensive, considering that this would also iterate over all non-map properties

dmonad (changes requested):
I can't accept this PR because of the performance overhead of iterating over all deleted structs and deleting them again.

I suggest that you look into how observer events on keys are generated. If `transaction.deleteSet` contains item.prev, then item.prev is deleted because we inserted something. Then the prev.item is the prevValue. Otherwise, the previous value was undefined.

WebCoder49:
Sorry about this. I am bearing your suggestions in mind now.

tests/y-map.tests.js
```diff
@@ -530,6 +530,253 @@ export const testChangeEvent = tc => {
...
+  map0.set('map', nmap)
+  map0.unobserve(setObserver)
+  map0.unobserveDeep(setDeepObserver)
+
+  const transactObserver = (/** @type {any} */e) => {
+    console.log("Transact observer called")
+    const keyChange = e.changes.keys.get('map')
+    t.assert(e.changes !== null && keyChange.action === 'delete' &&
+    !keyChange.oldValue.has('x') && keyChange.oldValue.get('map') == nnmap
+    && keyChange.oldValue.get('map').get('z') == 2 && keyChange.oldValue.get('y') == 1)
```
WebCoder49, line 577: I believe my assertion here should evaluate to false rather than true as my code currently makes it, since the `oldValue` should *not* contain "y" or "z" keys.

I ignored that the attribute-setting operations for "y" and "z" are also in the same transaction below. `YEvent.get keys()` shows that `oldValue` should contain the value **before the transaction** of the element at the key. Therefore, if that element's a map, by extension `oldValue` should contain the delete-states **before the transaction** of all fields of that element. Before the transaction, "y" and "z" keys had not been set on `nmap1`.

WebCoder49:
I've addressed the changes in a rewrite of the PR: #788

PR #789, opened 2026-07-18
docs(readme): add Deln0r/ygo to the Ports list

Deln0r:
Follow-up to yjs/docs#80, which you merged to add Deln0r/ygo to the docs ports page. The main README's Ports section still lists only reearth's port under "ygo", so this adds mine alongside it, disambiguated the same way the docs page does (ygo (Deln0r/ygo)).

Deln0r/ygo is a pure-Go port, byte-for-byte wire-compatible with the JS reference in V1 and V2, verified bidirectionally by cross-language fixtures. One line, mirroring the existing entry's format. Happy to reword or drop it if you would rather keep a single Go entry in the README.


dmonad:
thanks!

PR #795, opened 2026-08-03
Fix undo restoration with unreachable right origins

### LINKED ISSUE #794: Undo can restore sequence items with unreachable right origins
Undo restoration can select both a left and right origin that no longer define a reachable insertion interval after nested sequence edits. The restored item can then encode a right origin that is not reachable from its left origin, which can break fresh decoding or peer convergence.

This affects both the current main branch and v13.

I have patches with regression coverage for nested sequence undo round trips, fresh serialization, and collaborator convergence. The main-branch coverage also exercises a snapshot-rendered insertion. The fix verifies that the selected right origin remains reachable by walking rightward from the selected left origin; when it is not, restoration falls back to the immediate right neighbor before integrating the item.

I will send separate PRs for main and v13 that reference this issue.

sibnerian-oai:
Fixes #794.

When undo restores a sequence item after nested edits, the remembered right origin may no longer be reachable from the chosen left origin. Integrating that pair writes an invalid insertion interval, so a fresh decoder or collaborating peer can observe a different order.

This checks reachability before creating the restored item. If the remembered right origin is no longer on the left item right chain, restoration uses the immediate right neighbor instead.

The regression coverage exercises nested sequence undo round trips, fresh serialization, collaborator convergence, and a snapshot-rendered insertion.

Tests:
- `npm test`
- `npm run lint`

### DIFF
--- a/src/utils/UndoManager.js
+++ b/src/utils/UndoManager.js
@@ -512,6 +512,15 @@ export const redoItem = (transaction, item, redoitems, itemsToDelete, ignoreRemo
       left = parentType._map.get(item.parentSub) || null
     }
   }
+  if (left !== null && right !== null) {
+    let next = left.right
+    while (next !== null && next !== right) {
+      next = next.right
+    }
+    if (next === null) {
+      right = left.right
+    }
+  }
   const nextClock = store.getClock(ownClientID)
   const nextId = createID(ownClientID, nextClock)
   const redoneItem = new Item(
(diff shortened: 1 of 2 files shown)


dmonad:
```
if (left !== null && right !== null) {
    let next = left.right
    while (next !== null && next !== right) {
      next = next.right
    }
    if (next === null) {
      right = left.right
    }
  }
```

This looks wrong. So if it does not find right (which shouldn't happen) it simply sets `right = left.right` . Either we fix the non-existence of right (assume this was caused by redoing a deletion of a parent) or we always set `right = left.right` . Doing both just hides errors and will cause more trouble later (only one approach can be correct).

I don't have time to dig into this right now. Could you roughly explain how you produced this issue?

PR #800, opened 2026-08-24
fix: prevent throwing event handlers from breaking observer chain (fixes #799)

### LINKED ISSUE #799: A throwing event handler permanently breaks future document events and observers

**Checklist**

**Describe the bug**
If a listener for any of the following events throws, listeners and observers stop working for the remaining lifetime of the document: `beforeAllTransactions`, `beforeTransaction`, `afterTransactionCleanup`, `update`, `updateV2`, `subdocs`

**To Reproduce**
```
import * as Y from '@y/y';

const doc = new Y.Doc();
const type = doc.get('a');

let pushCount = 0;
function push() {
  pushCount++;
  console.log(`Pushing ${pushCount}`);
  type.push([pushCount]);
}

let firstObserveCount = 0;
type.observe(() => {
  firstObserveCount++;
});

// Attach an update listener that throws
const listener = () => {
  console.log('Throwing from update listener');
  throw new Error('provider error');
};
doc.on('update', listener);

// Write to the document and trigger the throwing listener
try {
  push();
} catch (error) {
  console.log('write threw:', error.message);
}

// Detach the throwing listener, document is permanently broken either way
doc.off('update', listener);

let secondObserveCount = 0;
type.observe(() => {
  secondObserveCount++;
});
let secondUpdateCount = 0;
doc.on('update', () => {
  secondUpdateCount++;
});

push();
push();
push();

console.log();
console.log(`Expected first observer to be called ${pushCount} times, got: ${firstObserveCount}`);
console.log(`Expected second observer to be called ${pushCount - 1} times, got: ${secondObserveCount}`);
console.log(`Expected second update listener to be called ${pushCount - 1} times, got: ${secondUpdateCount}`);
console.log('Array content:', type.toArray());
```

**Expected behavior**
See output of minimal reproduction

**Screenshots**
N/A

**Environment Information**
 - Only tested with node, but likely applicable everywhere else
 - Yjs version and the versions of the y-* modules you are using: @y/y@14.0.0-rc.24, also present in YJS v13 stable release

**Additional context**
N/A

* [ ] I'm a sponsor 💖
* [ ] This issue is a blocker for my project.

dikshit-n:
## Summary
Prevent a single throwing event listener from permanently breaking all future document events and observers.

## Problem
When any listener attached to `update`, `updateV2`, `beforeAllTransactions`, `beforeTransaction`, `afterTransactionCleanup`, `afterTransaction`, or `subdocs` throws an error, the transaction aborts mid-cleanup. The `afterAllTransactions` emit never fires, `doc._transactionCleanups` is left in a stale state, and all subsequent transactions silently fail — observers never fire again.

See: https://github.com/yjs/yjs/issues/799

### Fix 1 — `src/utils/EventHandler.js`
Replace the uncaught `callAll` loop with an explicit `for` loop wrapped in try-catch. Each listener is called in its own try block. If one throws, the error is saved and re-thrown after all remaining listeners have been called. This ensures all observers receive the event even when a sibling listener misbehaves.

### Fix 2 — `src/utils/Transaction.js`
Wrap each `doc.emit(...)` call in the `cleanupTransactions` finally block with a dedicated try-catch. Previously, a throwing listener on `update` would abort the entire finally block, preventing `afterAllTransactions` from ever firing and leaving `doc._transactionCleanups` stale. Now each emit is isolated — one throwing listener cannot prevent subsequent emits from firing.

## Changes Made
- `src/utils/EventHandler.js`: `callEventHandlerListeners` now uses an explicit loop with per-listener try-catch instead of `lib0/callAll`
- `src/utils/Transaction.js`: Each `doc.emit(...)` in the `finally` block of `cleanupTransactions` is wrapped in try-catch
- `tests/doc.tests.js`: Added `testThrowingEventHandlerDoesNotBreakObservers` regression test

## Testing
- Unit test added: `testThrowingEventHandlerDoesNotBreakObservers` in `tests/doc.tests.js`
  - Attaches a throwing `update` listener
  - Calls `transact` (throws, caught)
  - Removes the throwing listener
  - Adds a new observer and calls `transact` again
  - Asserts the new observer fires correctly and doc content is intact
- Full test suite: all 307 tests pass
- Linter: `standard` + `tsc` + `dpdm` all pass

## Checklist
- [x] Tests pass locally (307/307)
- [x] Lint passes (`standard`, `tsc`, `dpdm`)
- [x] Code follows repo style
- [x] No console.log or debug code left
- [x] No unrelated changes

### DIFF
--- a/src/utils/EventHandler.js
+++ b/src/utils/EventHandler.js
@@ -1,5 +1,3 @@
-import * as f from 'lib0/function'
-
 /**
  * General event handler implementation.
  *
@@ -75,6 +73,11 @@ export const removeAllEventHandlerListeners = eventHandler => {
  * Call all event listeners that were added via
  * {@link EventHandler#addEventListener}.
  *
+ * Each listener is called in a try-catch. If one throws, the error is
+ * re-thrown after all remaining listeners have been called. This ensures
+ * that a single throwing listener cannot permanently break the observer
+ * chain — all other listeners still receive the event.
+ *
  * @template ARG0,ARG1
  * @param {EventHandler<ARG0,ARG1>} eventHandler
  * @param {ARG0} arg0
@@ -83,5 +86,17 @@ export const removeAllEventHandlerListeners = eventHandler => {
  * @private
  * @function
  */
-export const callEventHandlerListeners = (eventHandler, arg0, arg1) =>
-  f.callAll(eventHandler.l, [arg0, arg1])
+export const callEventHandlerListeners = (eventHandler, arg0, arg1) => {
+  const listeners = eventHandler.l
+  let err = null
+  for (let i = 0; i < listeners.length; i++) {
+    try {
+      listeners[i](arg0, arg1)
+    } catch (e) {
+      err = e
+    }
+  }
+  if (err !== undefined) {
+    throw err
+  }
+}
--- a/src/utils/Transaction.js
+++ b/src/utils/Transaction.js
@@ -312,19 +312,28 @@ const cleanupTransactions = (transactionCleanups, i) => {
         doc.clientID = generateNewClientId()
       }
       // @todo Merge all the transactions into one and provide send the data as a single update message
-      doc.emit('afterTransactionCleanup', [transaction, doc])
+      // Each emit is wrapped in try-catch so a throwing listener on one event does not prevent
+      // subsequent emits from firing. This prevents the observer chain from being permanently
+      // broken by a single misbehaving listener.
+      try {
+        doc.emit('afterTransactionCleanup', [transaction, doc])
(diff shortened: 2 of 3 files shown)


dmonad:
Yjs does not allow errors in those event handlers. If you allow errors, you have to catch them yourself.

My personal opinion is that things should fail big-time when they fail. Otherwise, users won't notice. Currently, there is no proper way to rethrow in update / afterTransaction handlers. Simply catching an error (not even logging it, or rethrowing it) is an anti-pattern. It will simply hide bugs in your code. 

`callAll` already does what you are expecting. I don't think you would replace it. The issue is with the events that don't use the `callAll` approach.

